PR #7, opened 2021-11-09
Get authorities from GRANDPA or parachain staking (collators) in the dkg-metadata

Wizdave97:


pallets/dkg-metadata/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+	use frame_support::pallet_prelude::*;
+	use frame_system::pallet_prelude::*;
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		/// Authority identifier type
+		type DKGId: Member + Parameter + RuntimeAppPublic + Default + MaybeSerializeDeserialize;
+
+		/// Listener for authority set changes
+		type OnAuthoritySetChangeHandler: OnAuthoritySetChangeHandler<
```
Wizdave97, line 55: Thinking if we want to plug multiple pallets to receive changes to the authority set, we could make this a Tuple, but this works for now

PR #9, opened 2021-11-11
Write tests for proposal pallet and DKG authorities interaction

Wizdave97:


drewstone:
Can you document what these tests are doing and what tests you want to complete here?

pallets/dkg-proposals/src/tests.rs
```diff
@@ -517,3 +521,78 @@ fn proposal_expires() {
 		})]);
 	})
 }
+
+#[test]
+fn should_get_initial_proposers_from_dkg() {
```
Wizdave97, line 526: This test checks if the proposers are set after the DKG authorities are initialized in the genesis session

drewstone: Can we add tests to validate that they change when sessions update?

PR #10, opened 2021-11-16
submit validation

temld4:


pallets/dkg-proposal-handler/src/lib.rs
```diff
@@ -57,13 +57,41 @@ pub mod pallet {
...
+		pub fn submit_signed_proposal(
+			origin: OriginFor<T>,
+			prop: T::Proposal,
+		) -> DispatchResultWithPostInfo {
+			let encoded_proposal = prop.encode();
+			let eth_transaction = match TransactionV2::decode(&mut &encoded_proposal[..]) {
+				Ok(tx) => tx,
+				Err(_) => return Err(Error::<T>::ProposalFormatInvalid)?,
+			};
+			if let Err(err) = Self::validate_proposal_exists(&prop, &eth_transaction) {
```
drewstone, line 85: Would use ensure! Macro in place of this

pallets/dkg-proposal-handler/src/lib.rs
```diff
@@ -57,13 +57,41 @@ pub mod pallet {
...
+				return Err(err)?
+			}
+			if let Err(err) = Self::validate_ethereum_tx_signature(&eth_transaction) {
```
drewstone, line 88: Would use ensure! instead of this

pallets/dkg-proposal-handler/src/lib.rs
```diff
@@ -91,7 +121,49 @@ impl<T: Config> Pallet<T> {
...
+			},
+			TransactionV2::EIP1559(tx) => {
+				let chain_id: u64 = tx.chain_id;
+				let nonce = tx.nonce.as_u64();
+				let hash = EIP1559TransactionMessage::from(tx.clone()).hash();
+				(chain_id, nonce, hash)
+			},
+		};
+
+		if !pallet_dkg_proposals::Pallet::<T>::proposal_exists(chain_id, nonce, prop.clone()) {
```
drewstone, line 159: Ensure! works better than this pattern

drewstone (changes requested):
It seems like only the validation is handled. Do you plan to ha for storage of the proposals in a separate PR? I would suggest handling it all here. If the spec is clear let's chat because this is taking a bit of time.

pallets/dkg-proposal-handler/src/lib.rs
```diff
@@ -57,13 +57,41 @@ pub mod pallet {
...
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
 
 	#[pallet::call]
-	impl<T: Config> Pallet<T> {}
+	impl<T: Config> Pallet<T> {
+		#[pallet::weight(0)]
+		pub fn submit_signed_proposal(
+			origin: OriginFor<T>,
+			prop: T::Proposal,
```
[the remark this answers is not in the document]
drewstone: Consider this or using an enum to represent more Proposal states ```suggestion
			prop: T::SignedProposal,
```

File: pallets/dkg-proposals/src/lib.rs [hunk 5]
```diff
@@ -431,6 +431,27 @@ pub mod pallet {
...
+	fn proposal_exists(chain_id: u64, nonce: DepositNonce, prop: T::Proposal) -> bool {
+		let src_id = match T::ChainId::try_from(chain_id) {
+			Ok(v) => v,
+			Err(_) => return false,
+		};
+		return Votes::<T>::contains_key(src_id, (nonce, prop))
+	}
+
+	/// Removes specified proposal from the queue
+	fn remove_proposal(chain_id: u64, nonce: DepositNonce, prop: T::Proposal) -> bool {
```
drewstone, line 445: Why not make `chain_id` the proper type? 
```suggestion
	fn remove_proposal(chain_id: T::ChainId, nonce: DepositNonce, prop: T::Proposal) -> bool {
```

drewstone (changes requested):
We could add more metadata to the proposal type and make it easier to identify what type of tx we're working with. Consider using an enum here too:
```
pub enum ProposalType {
	EVMUnsigned {
		data: Vec<u8>,
	},
	EVMSigned {
		data: Vec<u8>,
		signature: Vec<u8>
	},
}

PR #13, opened 2021-11-18
Validate and submit proposal

temld4:


pallets/dkg-proposal-handler/src/lib.rs
```diff
@@ -117,15 +177,52 @@ impl<T: Config> Pallet<T> {
...
-		return match sp_io::crypto::secp256k1_ecdsa_recover(&sig, &msg) {
-			Ok(_) => Ok(()),
-			Err(_) => Err(DispatchError::BadOrigin),
-		}
+		return sp_io::crypto::secp256k1_ecdsa_recover(&sig, &msg).is_ok()
+	}
+
+	fn validate_proposal_signature(data: &Vec<u8>, signature: &Vec<u8>) -> bool {
+		let mut sig = [0u8; 65];
+		sig[0..64].copy_from_slice(&signature[..]);
```
Wizdave97, line 191: Shouldn't we check the length of the signature before making this copy

pallets/dkg-proposal-handler/src/lib.rs
```diff
@@ -117,15 +177,52 @@ impl<T: Config> Pallet<T> {
...
 
-		return match sp_io::crypto::secp256k1_ecdsa_recover(&sig, &msg) {
-			Ok(_) => Ok(()),
-			Err(_) => Err(DispatchError::BadOrigin),
-		}
+		return sp_io::crypto::secp256k1_ecdsa_recover(&sig, &msg).is_ok()
+	}
+
+	fn validate_proposal_signature(data: &Vec<u8>, signature: &Vec<u8>) -> bool {
+		let mut sig = [0u8; 65];
```
Wizdave97, line 190: Can we use a constant like `PROPOSAL_SIGNATURE_LENGTH` here instead of using just `65`

drewstone (approved):

PR #16, opened 2021-11-19
Have queued authorities partake in the MP keygen process before the next session

Wizdave97:


drewstone (commented):
Looks great so far!

Wizdave97:
I Still need to do a lot of testing on this, but can still be reviewed.

dkg-gadget/src/worker.rs
```diff
@@ -440,49 +499,93 @@ where
...
+				self.gossip_engine.lock().gossip_message(
+					dkg_topic::<B>(),
+					encoded_dkg_message.clone(),
+					true,
+				);
+				trace!(target: "dkg", "🕸️  Sent DKG Message {:?}", encoded_dkg_message);
+			}
+		};
+
+		if self.current_validator_set.id == GENESIS_AUTHORITY_SET_ID {
```
temld4, line 582: We should probably change this condition, since current dkg rounds need always send their outgoing messages for voting process. Maybe something like this?
`if let Some(id) =
			self.key_store.authority_id(self.current_validator_set.authorities.as_slice())
		{
			debug!(target: "dkg", "🕸️  Local authority id: {:?}", id);
			send_messages(&mut self.rounds, id);
		} else {
			panic!("error");
		}
		if self.current_validator_set.id != GENESIS_AUTHORITY_SET_ID {
			if let Some(id) =
				self.key_store.authority_id(self.queued_validator_set.authorities.as_slice())
			{
				debug!(target: "dkg", "🕸️  Local authority id: {:?}", id);
				if let Some(next_rounds) = self.next_rounds.as_mut() {
					send_messages(next_rounds, id);
				}
			} else {
				panic!("error");
			}
		}`

dkg-gadget/src/worker.rs
```diff
@@ -492,15 +595,34 @@ where
 	) {
 		debug!(target: "dkg", "🕸️  Process DKG message {}", &dkg_msg);
 
-		match self.rounds.handle_incoming(dkg_msg.payload) {
-			Ok(()) => (),
-			Err(err) => debug!(target: "dkg", "🕸️  Error while handling DKG message {:?}", err),
-		}
-		self.send_outgoing_dkg_messages();
+		if self.current_validator_set.id == GENESIS_AUTHORITY_SET_ID {
```
temld4, line 611: Same as above, we may receive incoming dkg message for any of the rounds or next_rounds. I think we can pass the same message to both of the rounds instances and let them decide whether they're interested in that particular message. (We'll need to add additional checks in MultiPartyECDSARounds::vote() method)

pallets/dkg-metadata/src/lib.rs
```diff
@@ -82,8 +101,50 @@ pub mod pallet {
...
+	pub type NextPublicKey<T: Config> =
+		StorageValue<_, (dkg_runtime_primitives::AuthoritySetId, Vec<u8>), ValueQuery>;
+
+	/// Holds active public key for ongoing session
+	#[pallet::storage]
+	#[pallet::getter(fn active_public_key)]
+	pub type ActivePublicKey<T: Config> =
+		StorageValue<_, (dkg_runtime_primitives::AuthoritySetId, Vec<u8>), ValueQuery>;
+
+	/// Holds public key for immedaiate past session
```
drewstone, line 142:
```suggestion
	/// Holds public key for immediate past session
```

PR #17, opened 2021-11-23
DKG Keygen refresh process

Wizdave97:


dkg-primitives/src/types.rs
```diff
@@ -80,12 +80,15 @@ pub struct DKGSignedPayload<Key> {
 	pub payload: Vec<u8>,
 	/// Signature for the payload
 	pub signature: Vec<u8>,
+	/// Bytes representation of the signature recid
+	pub signature_recid: Vec<u8>,
```
temld4, line 84: It's probably not necessary to have signature_recid, since we already have signature field to contain signature's bytes?

temld4 (approved):
Looks good!

dkg-primitives/src/types.rs
Wizdave97: The signature field is already converted to the runtime version of ecdsa, but in one of the tests we still needed the actual SignatureRecid, that's why I added them both

pallets/dkg-metadata/src/lib.rs
```diff
@@ -306,34 +389,139 @@ impl<T: Config> Pallet<T> {
...
+		if let Some(pub_key) = Self::next_dkg_public_key() {
+			let recovered_pub_key = match dkg_runtime_primitives::utils::recover_ecdsa_pub_key(
+				&pub_key.1, &signature,
+			) {
+				Ok(key) => key,
+				Err(_) => Err(Error::<T>::InvalidSignature)?,
+			};
+
+			if recovered_pub_key != Self::dkg_public_key().1 {
+				// TODO probably a slashing condition
```
drewstone, line 486: we need to also pass in the sender if we want to slash this.

drewstone (approved):
LGTM, some small things to address or provide docs on but otherwise a good step forward.

Wizdave97:
#19  needs to be merged before this, the DKGAPI was upated in this PR, so the changes made in the standalone node #19 will need to be updated here also.

Wizdave97:
> #19 needs to be merged before this, the DKGAPI was upated in this PR, so the changes made in the standalone node #19 will need to be updated here also.

Updated

PR #18, opened 2021-11-23
add proposal signing

temld4:


shekohex (approved):
LGTM!

standalone/runtime/src/lib.rs
```diff
@@ -457,6 +458,15 @@ impl_runtime_apis! {
 		fn set_dkg_pub_key(key: Vec<u8>) -> () {
 
 		}
+
+		fn get_unsigned_proposals() -> Vec<(ProposalNonce, ProposalType)> {
+			// DKGProposalHandler::get_unsigned_proposals()
+			Vec::new()
+		}
+
+		fn add_signed_proposal(proposal: ProposalType) {
```
drewstone, line 467: We should remove this. Runtime APIs are just for making queries to state not for making state changes.

drewstone (changes requested):
Need to remove any runtime APIs where we think we're mutating state (adding things). This is not how blockchains work. You need to submit extrinsics to mutate state.

dkg-gadget/src/worker.rs
```diff
@@ -650,16 +487,73 @@ where
 		self.process_finished_rounds();
 	}
 
+	fn process_finished_rounds(&mut self) {
+		for finished_round in self.rounds.get_finished_rounds() {
+			self.handle_finished_round(finished_round);
+		}
+
+		if let Some(mut next_rounds) = self.next_rounds.take() {
```
Wizdave97, line 495: We might be able to remove this, I don't think we'll ever conduct any voting process with the `next_rounds` value

PR #19, opened 2021-11-24
Update standalone runtime with custom pallets

Wizdave97:


Wizdave97:
There's still an issue with the authority sets at genesis, debugging that

shekohex (approved):

PR #20, opened 2021-11-24
fix proposal submission

temld4:


pallets/dkg-proposal-handler/src/lib.rs
```diff
@@ -135,35 +176,57 @@ impl<T: Config> Pallet<T> {
...
-		if let Ok(eth_transaction) = TransactionV2::decode(&mut &data[..]) {
-			ensure!(
-				Self::validate_ethereum_tx(&eth_transaction),
-				Error::<T>::ProposalFormatInvalid
-			);
+	fn submit_signed_proposal_onchain(block_number: T::BlockNumber) -> Result<(), &'static str> {
+		let submitted = StorageValueRef::persistent(b"dkg-proposal-handler::submitted");
 
-			let (chain_id, nonce) = Self::extract_chain_id_and_nonce(&eth_transaction)?;
+		let res = submitted.mutate(|submitted| match submitted {
```
Wizdave97, line 184: Since proposals submissions can happen anytime, why not remove the grace period check and submit any proposal found in the offchain storage, then clear the offchain storage after that, so tne check becomes  just this
``` 
				if let Ok(next_proposal) = Self::get_next_offchain_signed_proposal() {
					let _ = signer.send_signed_transaction(|_account| {
						Call::submit_signed_proposal { prop: next_proposal.clone() }
					});
				} ```

drewstone: What's the purpose of the GracePeriod? Can we document more often for this?

drewstone (approved):
Looks great pending understanding the grace period.

pallets/dkg-proposal-handler/src/lib.rs
Wizdave97: The grace period is like a spam protection, so we can throttle how many times an extrinsic is submitted by the offchain worker, though we might need a more elaborate solution here, since we want to prevent spamming, but we also want to submit signed proposals immediately they are available.

PR #21, opened 2021-11-25
Prevent duplicate submissions of public key signature and public key 

Wizdave97:


drewstone:
We should also ensure that when valid signatures are submitted that all other workers also clean their storage. Otherwise they may try to resubmit the public key signature. We need tests of this too.

PR #23, opened 2021-11-29
Aggregate public keys from keygen over gossip engine and batch submit on-chain

Wizdave97:


dkg-gadget/src/worker.rs
```diff
@@ -478,42 +493,168 @@ where
...
+	}
+
+	fn handle_public_key_broadcast(&mut self, dkg_msg: DKGMessage<Public, DKGPayloadKey>) {
+		if !self.dkg_state.listening_for_pub_key {
+			return
+		}
+
+		match dkg_msg.payload {
+			DKGMsgPayload::PublicKeyBroadcast(msg) => {
+				let key_and_sig = (msg.pub_key, msg.signature);
```
drewstone, line 589: We have to validate that this key is in fact a DKG authority / validator / collator. We run the risk of being open to spoofing attacks from non-authorities on the gossip network.

dkg-gadget/src/worker.rs
```diff
@@ -478,42 +493,168 @@ where
...
+				if !self.aggregated_public_keys.keys_and_signatures.contains(&key_and_sig) {
+					self.aggregated_public_keys.keys_and_signatures.push(key_and_sig);
+
+					if let Some(latest_header) = self.latest_header.as_ref() {
+						let threshold = self.get_threshold(latest_header).unwrap() as usize;
+						let num_authorities = self.queued_validator_set.authorities.len();
+						let threshold_buffer = num_authorities.saturating_sub(threshold) / 2;
```
drewstone, line 596: can you describe what this buffer is and why its there?

dkg-gadget/src/worker.rs
```diff
@@ -478,42 +493,168 @@ where
...
+							}
+						}
+					}
+				}
+			},
+			_ => {},
+		}
+	}
+
+	fn generate_random_delay(&self) -> Option<<B::Header as Header>::Number> {
```
drewstone, line 630: Can we document this? What distribution does it follow / what range of values does the delay take?

drewstone (changes requested):
Looking good!

dkg-gadget/src/worker.rs
Wizdave97: I wanted to aggregate a little more public keys above the threshold value,

pallets/dkg-metadata/src/lib.rs
```diff
@@ -608,6 +612,10 @@ impl<T: Config> OneSessionHandler<T::AccountId> for Pallet<T> {
 	where
 		I: Iterator<Item = (&'a T::AccountId, T::DKGId)>,
 	{
+		let changed = changed;
+		#[cfg(test)]
```
drewstone, line 616: This is really bad design. We shouldn't add such config flags into a runtime module like this.

pallets/dkg-metadata/src/lib.rs
```diff
@@ -351,7 +351,11 @@ impl<T: Config> Pallet<T> {
 		// As in GRANDPA, we trigger a validator set change only if the the validator
 		// set has actually changed.
 
-		if new != Self::authorities() {
+		let changed = new != Self::authorities();
+		#[cfg(test)]
```
drewstone, line 355: Here too, this should be avoided. It could potentially cause deviation in runtime binaries.

dkg-gadget/src/worker.rs
```diff
@@ -782,13 +799,13 @@ where
...
 					}
 				},
-				notification = self.block_import_notification.next().fuse() => {
-					if let Some(notification) = notification {
-						self.handle_import_notifications(notification);
-					} else {
-						return;
-					}
-				},
+				// notification = self.block_import_notification.next().fuse() => {
```
drewstone, line 802: Why is this commented out? Why not just remove it if its not being used, though shouldn't we be listening to finality notifications to start certain processes or no?

drewstone (changes requested):

PR #24, opened 2021-11-30
Tests for dkg-proposal-handler

temld4:


pallets/dkg-proposal-handler/src/tests.rs
```diff
@@ -1,7 +1,191 @@
...
+#[test]
+fn submit_signed_proposal_onchain_success() {
+	execute_test_with(|| {
+		let tx_v_2 = TransactionV2::EIP2930(mock_eth_tx_eip2930(0));
+
+		assert_ok!(DKGProposalHandler::handle_proposal(tx_v_2.encode(), ProposalAction::Sign(0)));
+		let unsigned_prop = DKGProposalHandler::unsigned_proposals(0, 0);
+		assert_eq!(unsigned_prop.is_some(), true);
+
+		add_proposal_to_offchain_storage(unsigned_prop.unwrap());
```
drewstone, line 102: I'm not sure I understand this test. Can we mock the intended flows more closely? If we submit signed value it should be signed no?

pallets/dkg-proposal-handler/src/tests.rs
```diff
@@ -1,7 +1,191 @@
...
+
+		assert_ok!(DKGProposalHandler::submit_signed_proposal_onchain(0));
+
+		check_offchain_proposals_num_eq(0);
+	});
+}
+
+#[test]
+fn submit_signed_proposal_success() {
```
drewstone, line 111: This looks like the correct version of the previous test. What are the differences between the two?

pallets/dkg-proposal-handler/src/tests.rs
```diff
@@ -1,7 +1,191 @@
...
+		));
+
+		assert_eq!(DKGProposalHandler::unsigned_proposals(0, 0).is_none(), true);
+		assert_eq!(DKGProposalHandler::signed_proposals(0, 0).is_some(), true);
+
+		// Second submission
+		assert_ok!(DKGProposalHandler::handle_proposal(tx_v_2.encode(), ProposalAction::Sign(0)));
+		assert_eq!(DKGProposalHandler::unsigned_proposals(0, 0).is_some(), true);
+
+		assert_ok!(DKGProposalHandler::submit_signed_proposal(Origin::root(), signed_proposal));
```
drewstone, line 150: Shouldn't this fail by the name of the test? It says fail if already exists?

pallets/dkg-proposal-handler/src/tests.rs
```diff
@@ -1,7 +1,191 @@
...
+
+		assert_ok!(DKGProposalHandler::handle_proposal(tx_v_2.encode(), ProposalAction::Sign(0)));
+		assert_eq!(DKGProposalHandler::unsigned_proposals(0, 0).is_some(), true);
+
+		let mut invalid_sig: Vec<u8> = Vec::new();
+		invalid_sig.extend_from_slice(&[0u8, 64]);
+		let signed_proposal =
+			ProposalType::EVMSigned { data: tx_v_2.encode(), signature: invalid_sig };
+
+		assert_eq!(
```
drewstone, line 183: This seems like it should also fail due to invalid sig.

drewstone (changes requested):
Looks like some tests need addressing or are still WIP. It would be great for integration testing if we can ensure that the signed txes are submittable on a local Ethereum network. You should spin up a ganache instance with these tests to go through a full integration flow.

PR #25, opened 2021-12-07
Update substrate dependecies to master

Wizdave97:


Wizdave97:
Was having clashes with a dependency I wanted to use in the integraton tests, that's why I decided to update this to master

PR #26, opened 2021-12-07
Keygen refresh integration tests

Wizdave97:


dkg-test-suite/src/utils.ts
```diff
@@ -1,19 +1,42 @@
...
-exports.listenOneBlock = async function (api) {
+export const listenOneBlock = async function (api: ApiPromise) {
     const unsubscribe = await api.rpc.chain.subscribeNewHeads((header) => {
         console.log(`Chain is at block: #${header.hash}`);
         unsubscribe();
     });
 }
 
-exports.printValidators = async function (api) {
+export const wait_n_finalized_blocks = async function (api: ApiPromise, n: number, timeout:  number) {
```
drewstone, line 22: We should use camelCase for TS

dkg-test-suite/src/key_refresh.ts
```diff
@@ -0,0 +1,128 @@
...
+
+	const keyring = new Keyring({ type: "sr25519" });
+	const alice = keyring.addFromUri("//Alice");
+	const charliesStash = keyring.addFromUri("//Charlie//stash");
+
+	// Session 0
+	console.log("Session 0");
+
+	let forceNewEra = api.tx.staking.forceNewEraAlways();
+
```
drewstone, line 16: too many new lines, condense this a bit if you can.

drewstone (changes requested):

Wizdave97:
The master branch of substrate has been updated and the Default trait has been removed from the AccountId,  The only pallet left to be fixed is the parachain staking pallet, I'll fix that over the weekend, the tests are and bug fixes are done for the refresh though.

PR #27, opened 2021-12-09
Use sr25519 to sign extrinsics  and submit genesis authority set public key on chain

Wizdave97:


pallets/dkg-metadata/src/lib.rs
```diff
@@ -600,8 +600,9 @@ impl<T: Config> Pallet<T> {
 				Ok(key) => key,
 				Err(_) => Err(Error::<T>::InvalidSignature)?,
 			};
+			let stored_key = &Self::dkg_public_key().1;
 
-			if recovered_pub_key != Self::dkg_public_key().1 {
+			if recovered_pub_key != stored_key[1..].to_vec() {
```
drewstone, line 605: Can you add docs to this? Why index at 1?

drewstone (changes requested):

drewstone (changes requested):

Wizdave97:
There's only on issue left here that I'm still debugging, when the refresh voting starts, it does not progress, so it's never completed, once that is fixed, this would be ready to go.

Wizdave97:
> There's only on issue left here that I'm still debugging, when the refresh voting starts, it does not progress, so it's never completed, once that is fixed, this would be ready to go.

Fixed up

Wizdave97:
https://user-images.githubusercontent.com/31099392/145551452-1c883e70-103e-4ca0-b2c5-b2946cd51240.mp4

shekohex (approved):

PR #28, opened 2021-12-14
Add Support for Update Anchor Proposal Over DKG

shekohex:


pallets/dkg-proposal-handler/src/lib.rs
```diff
@@ -116,34 +121,98 @@ pub mod pallet {
 			_origin: OriginFor<T>,
 			prop: ProposalType,
 		) -> DispatchResultWithPostInfo {
+			// let sender = ensure_signed(origin)?;
```
drewstone, line 124: Should there be a TODO here? Is there anything wrong with having the origin be any `Signed` origin?

drewstone (approved):

pallets/dkg-proposal-handler/src/lib.rs
shekohex: Nope, it is good, as we check for the data and signature but we don't want to need the check anyway.
We can uncomment it.

Wizdave97: We should also check the signature to see if the signer is a dkg authority, that's not been done currently.

shekohex: Yes, that's correct! we should check that against the current dkg public key!

drewstone: If we don't check that it is signed shady then anyone can call this and we can't charge tx fees. That means it can be spammed potentially?

shekohex: @Wizdave97 could you create a task for that? so maybe someone could work on it?

standalone/node/src/chain_spec.rs
```diff
@@ -229,5 +229,31 @@ fn testnet_genesis(
 			threshold: Default::default(),
 			authority_ids: initial_authorities.iter().map(|(x, ..)| x.clone()).collect::<_>(),
 		},
+		dkg_proposals: DKGProposalsConfig {
+			initial_chain_ids: vec![
+				5001, // Hermis
+				5002, // Athena
+			],
+			initial_r_ids: vec![
+				// Resoruce ID for Chain Hermis => Athena
```
drewstone, line 238:
```suggestion
				// Resource ID for Chain Hermis => Athena
```

pallets/dkg-proposal-handler/src/lib.rs
drewstone, line 124:
```suggestion
			let _sender = ensure_signed(origin)?;
```

PR #31, opened 2021-12-16
Rerun keygen for queued validators if the set changes when authority set has not changed

Wizdave97:
When a session change occurs, the queued validator set can change even when the active authority set does not change, we were not taking care of such a situation before.


drewstone (approved):
LGTM

PR #32, opened 2021-12-18
New proposals for root origins (democracy system)

drewstone:
# Overview
This PR introduces two new proposal types.
- `TokenUpdateProposal`
- `WrappingFeeUpdateProposal`

## TokenUpdateProposal
This proposal is for adding tokens to a `GovernedTokenWrapper` or `SignatureGovernedTokenWrapper`'s valid token whitelists. This proposal should not be submittable by a `proposer`.

## WrappingFeeUpdateProposal
This proposal is for updating the wrapping fee of a `GovernedTokenWrapper` or `SignatureGovernedTokenWrapper`. The proposal should also not be submittable by a `proposer`.

The structure non-EVM tx proposals are aggregated under a `single-parameter` regime. For each type of proposal -- `AnchorUpdateProposal`, `TokenUpdateProposal`, or `WrappingFeeUpdateProposal` -- there is really only 1 parameter being updated: a 32-bytes merkle root, a token `address`, or a `uint`. All these proposals benefit from the same structure, since we also ought to leverage nonces to prevent any replay attacks on any given contract instance.

# Future work
In doing this, it's now clear that we should create a new `Handler` for the TokenWrapper contracts as well as use resource IDs and integrate the flow back into the bridge. It allows us to coordinate governance execution of what the DKG is required for at the `Bridge.sol` and `pallet-bridge` levels.

This would create a setup like:
- `GovernedTokenWrapper` with the `TokenWrapperHandler` as the governor/admin.
- A new `TokenWrapperHandler` that either updates the tokens set or the wrapping fee
- A resource ID that is `(address_of_token_wrapper, chain_id)` (i.e. matching the existing structure).


drewstone:
@saiakilesh this is relevant for some future work on the Solidity side as well. Check this out and lmk if it's clear what we should do next.

Wizdave97 (approved):

PR #43, opened 2021-12-23
Adds new resource ID proposal and function sig to all proposals

drewstone:
# Overview
This adds the resource ID proposal #42 to the DKG as well as updates all proposal decoding to the new structure

## Non-anchor proposal types
- Single parameter proposals
    - `(execution_chain_id, nonce, functionSig, newToken)`
    - `(execution_chain_id, nonce, functionSig, newFee)`
- Double parameter proposals
    - `(execution_chain_id, nonce, functionSig, param1, param2)`
- Triple parameter proposals
    - Setting resource ID, handler, execution contexts
    - `(execution_chain_id, nonce, functionSig, resourceId, handlerAddress, executionContextAddress)`

# Caveats & Issues
We need to think more here as we need to be sure we prevent a single proposal from being able to update multiple bridges. The execution_chain and nonce actually doesn't solve replay issues when we have multiple bridges and plan to deploy new ones.

For example, if we sign `(1, 1, keccak256(setFee(uint)), tokenAddress)`, while we may be protected from executing this _again_ on the same contract, it doesn't seem like it prevents anyone from replaying this on a _different_ GovernedToken if the governor is the same as the DKG. 

The same likely holds for anchor update proposals across totally different anchors on the same chain. Say we sign, `(chain_id, latest_leaf_index, merkle_root)`. After calling `executeProposalWithSig` with this data, we can't call it again on this anchor. But, it appears that we can call it on any other anchor whose latest leaf index is less than `latest_leaf_index`.

## Potential solution
Since we use resource IDs as contract addresses + chain_ids in a single `bytes32`, we can include this in our proposals. Then for every proposal, we check the chain ID matches the executing chain's chain identifier and guarantee that proposals are then specific to a given contract instance.

CC @saiakilesh @shekohex

### DIFF
--- a/pallets/dkg-proposal-handler/src/lib.rs
+++ b/pallets/dkg-proposal-handler/src/lib.rs
@@ -684,7 +684,7 @@ impl<T: Config> Pallet<T> {
 	fn decode_chain_id_nonce_from_proposal<const N: usize>(
 		data: &[u8],
 	) -> Result<(T::ChainId, ProposalNonce), Error<T>> {
-		// check if the data length is 118 bytes [
+		// check if the data length is (86 + N) bytes [
 		// 	handler_address_0x_prefixed(22),
 		// 	chain_id(32),
 		// 	nonce(32),
--- a/pallets/dkg-proposal-handler/src/lib.rs
+++ b/pallets/dkg-proposal-handler/src/lib.rs
@@ -688,7 +688,7 @@ impl<T: Config> Pallet<T> {
 		// 	handler_address_0x_prefixed(22),
 		// 	chain_id(32),
 		// 	nonce(32),
-		//  function_sig(4),
+		//  (function_sig(4) + parameters_length()) -> N,
 		// 	parameter(N)
 		// ]
 		if data.len() != 22 + 32 + 32 + N {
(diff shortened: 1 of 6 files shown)


pallets/dkg-proposal-handler/src/lib.rs
```diff
@@ -643,17 +661,34 @@ impl<T: Config> Pallet<T> {
...
+		// function signature + 3 parameters: (bytes32 resourceId, address handlerAddress, address executionContext)
+		// 4 + 32 + 32 + 32 = 100 bytes
+		Self::decode_chain_id_nonce_from_proposal::<100>(data)
 	}
 
-	fn decode_single_parameter_proposal<const N: usize>(
+	fn decode_chain_id_nonce_from_proposal<const N: usize>(
 		data: &[u8],
 	) -> Result<(T::ChainId, ProposalNonce), Error<T>> {
 		// check if the data length is 118 bytes [
```
Wizdave97, line 687:
```suggestion
		// check if the data length is (86 + N) bytes [
```

pallets/dkg-proposal-handler/src/lib.rs
```diff
@@ -643,17 +661,34 @@ impl<T: Config> Pallet<T> {
...
 		// 	handler_address_0x_prefixed(22),
 		// 	chain_id(32),
 		// 	nonce(32),
+		//  function_sig(4),
```
Wizdave97, line 691:
```suggestion
		//  (function_sig(4) + parameters_len()) -> N,
```

pallets/dkg-proposal-handler/src/lib.rs
```diff
@@ -643,17 +661,34 @@ impl<T: Config> Pallet<T> {
...
 		// 	parameter(N)
```
Wizdave97, line 692:
```suggestion
```

Wizdave97 (approved):

PR #44, opened 2021-12-23
add misbehaviour report creation

temld4:


dkg-gadget/src/worker.rs
```diff
@@ -442,8 +448,11 @@ where
 		debug!(target: "dkg", "🕸️  Try sending DKG messages");
 
 		let send_messages = |rounds: &mut MultiPartyECDSARounds<DKGPayloadKey>,
-		                     authority_id: Public| {
-			rounds.proceed();
+		                     authority_id: Public|
+		 -> Result<(), DKGError> {
+			if let Err(err) = rounds.proceed() {
```
drewstone, line 453: This pattern here too should be fixed.

drewstone (changes requested):

drewstone:
Merge conflicts here. Can you document how these misbehaviors can be detected on-chain? Are there proofs or do we need to develop a voting/tallying system for it.

PR #45, opened 2021-12-24
Persistent DKG Restarts

Wizdave97:


drewstone:
Merge conflicts from latest merge @Wizdave97 here now.

drewstone (changes requested):
Looks good! Would like more docs on how this is being tested and how you've tested this so far? Have you done a live test with nodes falling off during keygen?

Wizdave97:
> Looks good! Would like more docs on how this is being tested and how you've tested this so far? Have you done a live test with nodes falling off during keygen?

I've tested this with three live nodes 

The tests I've run is with a node joining the keygen late when it cannot receive all the messages sent by other peers anymore.
Also tested when a node drops off before the keygen completes.

temld4 (approved):

PR #51, opened 2021-12-29
Sign and verify gossip messages before sending / processing

drewstone:


dkg-gadget/src/worker.rs
```diff
@@ -588,6 +609,58 @@ where
...
+						}))
+					}).collect(),
+					&encoded,
+					&signature,
+				)
+				.1
+			};
+
+			if !check_signers(authority_accounts.clone().unwrap().0.into())
+				|| !check_signers(authority_accounts.clone().unwrap().1.into()) {
```
temld4, line 650: Shouldn't it be && instead of || ?

dkg-gadget/src/worker.rs
```diff
@@ -588,6 +609,58 @@ where
...
+			} else {
+				return Ok(dkg_msg);
+			}
+		};
+
+		if !can_proceed {
+			return Err("Message signature is not from a registered authority or next authority".into());
+		}
+
+		Err("Invalid sender: not an authority or next authority".into())
```
temld4, line 661: Could you explain how this statement may be reached? I'm not sure, but it feels like can_proceed can only be false, in other cases we return Ok or Err out of function.

dkg-gadget/src/worker.rs
drewstone: Yea this is fixed now, I want messages to be valid if they come from either current or next authorities. Double-check this looks good now @temld4 .

dkg-gadget/src/worker.rs
drewstone: Good point, I removed the need for can_proceed.

PR #59, opened 2022-01-05
Continuous offline stage restarts

Wizdave97:


dkg-gadget/src/worker.rs
```diff
@@ -1257,23 +1247,63 @@ where
...
+		let mut errors = vec![];
+		if rounds.is_offline_ready() {
+			for (key, ..) in &unsigned_proposals {
+				if self.dkg_state.reset_signers_at.contains_key(&key.encode()) {
+					continue
+				}
+				// TODO: use deterministic random signers set
+				let signer_set_id = rounds.get_id();
+				let s_l = (1..=rounds.dkg_params().2).collect();
+				match rounds.reset_signers(key.encode(), signer_set_id, s_l, *header.number()) {
```
drewstone, line 1272: More comments here necessary. Is this using the on-chain threshold of signers?

dkg-gadget/src/worker.rs
```diff
@@ -1257,23 +1247,63 @@ where
 		}
 	}
 
+	/// Get unsigned proposals and create offline stage using an encoded (chainId, dkgpayloadKey) as the round key
```
drewstone, line 1250:
```suggestion
	/// Get unsigned proposals and create offline stage using an encoded (ChainId, DKGPayloadKey) as the round key
```

dkg-gadget/src/worker.rs
```diff
@@ -1257,23 +1247,63 @@ where
...
+		let unsigned_proposals = match self.client.runtime_api().get_unsigned_proposals(&at) {
+			Ok(res) => res,
+			Err(_) => return,
+		};
+
+		let rounds = self.rounds.as_mut().unwrap();
+		let mut errors = vec![];
+		if rounds.is_offline_ready() {
+			for (key, ..) in &unsigned_proposals {
+				if self.dkg_state.reset_signers_at.contains_key(&key.encode()) {
```
drewstone, line 1266: Need comments here.

dkg-gadget/src/worker.rs
drewstone: Ensure we define whether we use:
- the same signing set for all offline stages within an epoch
- rotate it based on the `(ChainId, DKGPayloadKey)` deterministically

drewstone: Consider choosing the signing set once each time the authority set changes. Ensure that we are using the on-chain defined signing threshold.

standalone/runtime/src/lib.rs
```diff
@@ -832,7 +832,7 @@ impl_runtime_apis! {
 			return None
 		}
 
-		fn get_unsigned_proposals() -> Vec<(DKGPayloadKey, ProposalType)> {
+		fn get_unsigned_proposals() -> Vec<((u32, DKGPayloadKey), ProposalType)> {
```
drewstone, line 835: Can we use the ChainId type?

runtime/src/lib.rs
```diff
@@ -794,7 +794,7 @@ impl_runtime_apis! {
 			return None
 		}
 
-		fn get_unsigned_proposals() -> Vec<(DKGPayloadKey, ProposalType)> {
+		fn get_unsigned_proposals() -> Vec<((u32, DKGPayloadKey), ProposalType)> {
```
drewstone, line 797: Can we use the ChainId type?

drewstone (changes requested):

drewstone (approved):

PR #63, opened 2022-01-07
Update repo with new multi-party-ecdsa version

drewstone:


drewstone:
Fixed this @temld4 by not updating the polkadot/substrate/cumulus dependencies. Will post a task to update.

PR #82, opened 2022-01-19
Refactor DKG rounds

temld4:


temld4:
Changes to be made:
1) Make OfflineRounds work with single stage and not with map
2) Make SignRounds work with single round tracker instead of map
3) Test

dkg-primitives/src/rounds/dkg.rs
```diff
@@ -353,48 +353,39 @@ where
...
-										sig,
-										sign_manual
-									));
-							}
+		if let Some(OfflineState::Finished(Ok(completed_offline))) = self.offlines.remove(&round_key) {
+			let hash = BigInt::from_bytes(&keccak_256(&data));
+
+			match SignManual::new(hash, completed_offline.clone()) {
+				Ok((sign_manual, sig)) => {
+					trace!(target: "dkg", "🕸️  Creating vote /w key {:?}", &round_key);
```
drewstone, line 362: `Creating vote w/ key

dkg-gadget/src/persistence.rs
```diff
@@ -190,7 +188,6 @@ where
 
 				if queued_local_key.is_some() {
 					rounds.set_local_key(queued_local_key.as_ref().unwrap().local_key.clone());
-					rounds.set_stage(Stage::OfflineReady);
 					// // We set the signer set using the public key as a seed to select signers at random
```
shekohex, line 191: Remove the extra //

dkg-primitives/src/rounds/dkg.rs
```diff
@@ -0,0 +1,687 @@
...
+	gg_2020::state_machine::{keygen as gg20_keygen, sign as gg20_sign, traits::RoundBlame},
+};
+
+/// DKG State tracker
+pub struct DKGState<C>
+where
+	C: AtLeast32BitUnsigned + Copy,
+{
+	pub accepted: bool,
+	pub is_epoch_over: bool,
```
shekohex, line 30: Would be better to use `epoch_is_over` over `is_epoch_over` here

In general:
> Do name Boolean properties with an affirmative phrase (CanSeek instead of CantSeek). Optionally, you can also prefix Boolean properties with Is, Can, or Has, but only where it adds value.

dkg-primitives/src/rounds/dkg.rs
```diff
@@ -0,0 +1,687 @@
...
+	local_keystore: Option<Arc<LocalKeystore>>,
+}
+
+impl<C> MultiPartyECDSARounds<C>
+where
+	C: AtLeast32BitUnsigned + Copy,
+{
+	/// Public ///
+
+	pub fn new(
```
shekohex, line 76: Could we use builder pattern here? instead of this huge new method?

dkg-primitives/src/rounds/dkg.rs
```diff
@@ -0,0 +1,687 @@
+use curv::{arithmetic::Converter, elliptic::curves::Secp256k1, BigInt};
+use log::{error, info, trace, warn};
+
+use sc_keystore::LocalKeystore;
+use sp_core::sr25519;
+use sp_runtime::traits::AtLeast32BitUnsigned;
+use std::{collections::HashMap, path::PathBuf, sync::Arc};
+
+use super::{keygen::*, offline::*, sign::*};
+use std::mem::replace;
```
shekohex, line 10: would be better to import it as `use std::mem` so when we use it would read `mem::replace` insead of just `replace` so we know what exactly this function came from and what it is doing.

dkg-primitives/src/rounds/keygen.rs
```diff
@@ -0,0 +1,288 @@
...
+					let serialized = serde_json::to_string(&m).unwrap();
+					return DKGKeygenMessage { keygen_set_id, keygen_msg: serialized.into_bytes() }
+				})
+				.collect::<Vec<DKGKeygenMessage>>();
+
+			keygen.message_queue().clear();
+			return enc_messages
+		}
+
+		vec![]
```
shekohex, line 211: I'd better move this into an `else` block, and use `Vec::new()`.

dkg-primitives/src/rounds/keygen.rs
```diff
@@ -0,0 +1,288 @@
...
+		if data.keygen_set_id != self.params.keygen_set_id {
+			return Err(DKGError::GenericError { reason: "Keygen set ids do not match".to_string() })
+		}
+
+		let keygen = &mut self.keygen;
+
+		trace!(target: "dkg", "🕸️  Handle incoming keygen message");
+		if data.keygen_msg.is_empty() {
+			warn!(
+				target: "dkg", "🕸️  Got empty message");
```
shekohex, line 226: formatting here, this could be moved to the line above.

dkg-primitives/src/rounds/keygen.rs
```diff
@@ -0,0 +1,288 @@
...
+			return Ok(())
+		}
+
+		let msg: Msg<ProtocolMessage> = match serde_json::from_slice(&data.keygen_msg) {
+			Ok(msg) => msg,
+			Err(err) => {
+				error!(target: "dkg", "🕸️  Error deserializing msg: {:?}", err);
+				return Err(DKGError::GenericError {
+					reason: "Error deserializing keygen msg".to_string(),
```
shekohex, line 235: we could use `format!("Error deserializing keygen msg, reason: {}", e),` so we get the actual error message too.

dkg-primitives/src/rounds/keygen.rs
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	fn try_finish(mut self) -> Result<LocalKey<Secp256k1>, DKGError> {
+		info!(target: "dkg", "🕸️  Keygen is finished, extracting output, round_id: {:?}", self.params.round_id);
+		match self.keygen.pick_output() {
+			Some(Ok(key)) => {
+				info!(target: "dkg", "🕸️  local share key is extracted");
+				Ok(key)
+			},
+			Some(Err(e)) => panic!("Keygen finished with error result {}", e),
```
shekohex, line 284: is panicking here is desirable? I mean we return a result here, so we could just return another error, and it is up to the caller to handle these error instead of panicing here without any control.

dkg-primitives/src/rounds/offline.rs
```diff
@@ -0,0 +1,305 @@
...
+			trace!(target: "dkg", "🕸️  before: {:?}", offline_stage);
+			// TODO, handle asynchronously
+			match offline_stage.proceed() {
+				Ok(_) => {
+					trace!(target: "dkg", "🕸️  after: {:?}", offline_stage);
+				},
+				Err(err) => {
+					println!("Error proceeding offline stage {:?}", err);
+					match err {
+						gg20_sign::Error::ProceedRound(proceed_err) => match proceed_err {
```
shekohex, line 151: You can do the following here instead of matching on every single one of them you could match on all of them if they are all have the same inner struct.

```rust
match proceed_err {
  gg20_sign::rounds::Error::Round1(e)
  | gg20_sign::rounds::Error::Round2Stage4(e)
  | gg20_sign::rounds::Round3::Round3(e)
  | // and so on here
  => return Err(DKGError::OfflineMisbehaviour { bad_actors: vec_usize_to_u16(e.bad_actors), }),
}
```

dkg-primitives/src/rounds/sign.rs
```diff
@@ -0,0 +1,373 @@
...
+	round_key: Vec<u8>,
+	sign_tracker: DKGRoundTracker<Vec<u8>, Clock>,
+	sign_outgoing_msgs: Vec<DKGVoteMessage>,
+}
+
+impl<C> SignRounds<C>
+where
+	C: AtLeast32BitUnsigned + Copy,
+{
+	pub fn new(
```
shekohex, line 122: We could always use the builder pattern here

dkg-primitives/src/rounds/sign.rs
```diff
@@ -0,0 +1,373 @@
...
+					debug!("Obtained complete signature: {}", &sig.recid);
+					Ok(sig)
+				},
+				Err(err) => {
+					let sign_err = match err {
+						SignError::LocalSigning(sign_err) => sign_err,
+						SignError::CompleteSigning(sign_err) => sign_err,
+					};
+
+					match sign_err {
```
shekohex, line 306: same here you could collapse all these match cases into one.

dkg-primitives/src/rounds/sign.rs
```diff
@@ -0,0 +1,373 @@
...
+			}
+		}
+		Err(DKGError::GenericError { reason: "No SignManual found".to_string() })
+	}
+}
+
+pub fn convert_signature(sig_recid: &SignatureRecid) -> Option<Signature> {
+	let r = sig_recid.r.to_bigint().to_bytes();
+	let s = sig_recid.s.to_bigint().to_bytes();
+	let v = sig_recid.recid;
```
shekohex, line 339: Could this be `let v = sig_recid.recid + 27u8;` so it would be EVM compatible?

shekohex (changes requested):
Looks good, I've left a few comments, should be easy. Otherwise happy to merge this PR.

PR #123, opened 2022-01-25
Add historical refresh data to storage and update chain id/proposal type for EVM prefix signed data

drewstone:
Related: https://github.com/webb-tools/protocol-solidity/issues/108


drewstone:
I fixed the rotation but persistence does not work with it. It seems that the persistence system does not survive restarts. It fails to sign the next public key properly after nodes restart (tested by terminating the local testnet and starting the nodes again).

shekohex (approved):
LGMT! The Refresh Tests work without any issues now.
```
$ ts-node ./src/testDkgRefresh.ts
Round 1/3 ✔ Public key: 0x03287207b8cf6dabfceb072bb27b8e9a4d08bc006c5bbc88db3ec751c0d7be5e87
Round 2/3 ✔ Public key: 0x03ba4f7f8f6121a1c9fce801943805a2be16864ec35ab3fe20eb90fd6ba880b278
Round 3/3 ✔ Public key: 0x028279e36bc3e6ae666542651237b98e99f258ce7951b76d842524be9480f5e0ae
Done
```

PR #143, opened 2022-02-01
Fix persistence for key rotation

temld4:


dkg-gadget/src/persistence.rs
```diff
@@ -245,9 +247,8 @@ where
 						.to_bytes(true)[1..];
 					let set = (1..=rounds.dkg_params().2).collect::<Vec<_>>();
 					let signers_set = select_random_set(seed, set, rounds.dkg_params().1 + 1);
-					if let Ok(signers_set) = signers_set {
-						let round_id = rounds.get_id();
-						rounds.set_signer_set_id(round_id);
+					if let Ok(mut signers_set) = signers_set {
```
drewstone, line 250: Can we add comments as to what values are in this set? Is it authority ids or is it just party indices?

dkg-gadget/src/persistence.rs
```diff
@@ -207,9 +210,8 @@ where
 							[1..];
 					let set = (1..=rounds.dkg_params().2).collect::<Vec<_>>();
 					let signers_set = select_random_set(seed, set, rounds.dkg_params().1 + 1);
-					if let Ok(signers_set) = signers_set {
-						let round_id = rounds.get_id();
-						rounds.set_signer_set_id(round_id);
+					if let Ok(mut signers_set) = signers_set {
+						signers_set.sort();
```
drewstone, line 214: Can you describe the benefit of sorting here?

drewstone (changes requested):
Partial review, would be great to add more comments to the code describing everything.

PR #145, opened 2022-02-01
Dami/configurable limit proposals

dharjeezy:
Add a proposal for configuring limits of a VAnchor as described in #141

### DIFF
--- a/dkg-test-suite/src/testVAnchorConfigurableLimitsProposal.ts
+++ b/dkg-test-suite/src/testVAnchorConfigurableLimitsProposal.ts
@@ -25,6 +25,14 @@ async function testDirectProposal() {
 					'WrappingFeeUpdate',
 					'WrappingFeeUpdateSigned',
                     // TODO: Add proposals for configurable limits
+					'MaxDepositLimitUpdate',
+					'MaxDepositLimitUpdateSigned',
+					'MinWithdrawalLimitUpdate',
+					'MinWithdrawalLimitUpdateSigned',
+					'MaxExtLimitUpdate',
+					'MaxExtLimitUpdateSigned',
+					'MaxFeeLimitUpdate',
+					'MaxFeeLimitUpdateSigned'
 				],
 			},
 			EVMUnsigned: {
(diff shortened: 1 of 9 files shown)


pallets/dkg-proposal-handler/src/lib.rs
```diff
@@ -831,6 +948,74 @@ impl<T: Config> Pallet<T> {
...
+		to_bytes.copy_from_slice(&data[60..80]);
+		let to = Address::from(to_bytes);
+		let mut amount_to_rescue_bytes = [0u8; 32];
+		amount_to_rescue_bytes.copy_from_slice(&data[80..112]);
+		Ok(header)
+	}
+
+	/// (header: 40 Bytes, max_deposit_limit_bytes: 32) = 72
+	/// Bytes
+	fn decode_max_deposit_limit_proposal(
```
drewstone, line 969: Functions look great and they all do the same thing, can we aggregate the logic in one function for parameters with this byte lengths? Maybe using const generics or do you think it's a bad idea to de-deduplicate the code?

drewstone (changes requested):

dkg-test-suite/src/testVAnchorConfigurableLimitsProposal.ts
```diff
@@ -0,0 +1,241 @@
...
+import { assert } from '@polkadot/util';
+
+const raw_data =
+	'00000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000001';
+const tokenUpdateProp = new Uint8Array(hexToBytes(raw_data));
+
+async function testDirectProposal() {
+	const api = await ApiPromise.create({
+		provider,
+		types: {
```
drewstone, line 16: Can we move these types into a file and export them for these tests? Also these types should be updated in the other tests now that this PR introduces new types.

drewstone (changes requested):

pallets/dkg-proposal-handler/src/lib.rs
dharjeezy: Yh... I was thinking of making it a one function call for all but then I thought what if the byte length for each of them changes in the future, well I guess by then we can make the decoding specific. But, for now, I will make the function just one for all as you suggested.

dkg-test-suite/src/testVAnchorConfigurableLimitsProposal.ts
dharjeezy: Alright

PR #149, opened 2022-02-03
Address issue #74 and its blockers

### LINKED ISSUE #74: [BUG] Sign / keygen cache does not work
## Description
Persistent DKG key material doesn't work as expected when using 3-node testnet with `--tmp` DBs (and likely when nodes are separated, which is untested). We should identify how to test this more deeply and decide on the expected behavior.

## Steps to reproduce
1. Run the 3 node testnet, finish keygen.
2. Shut off a node and wait.
3. Turn the node back on.
4. Initiate anchor proposal sign script, proposal succeeds but never gets signed.

### LINKED ISSUE #146: [BUG] Submitting Signed Proposal fails due to an invalid header/data structure
## Overview

When the ocw tries to submit the signed proposal from the off-chain storage back into on-chain through an xt, it fails due to an invalid header/data structure, specifically speaking about the stage when it tries to verify that this signed proposal is not already in the on-chain storage.

if we take a look here, we see we filter out the signed proposal that is already on-chain:
https://github.com/webb-tools/dkg-substrate/blob/1e4e0e53f95cd2fed82aaf9e5dcefe1a91cdb2e0/pallets/dkg-proposal-handler/src/lib.rs#L540-L545

And the way we do so is through the `is_existing_proposal` and if we take a closer look, specifically at this match branch here:
https://github.com/webb-tools/dkg-substrate/blob/1e4e0e53f95cd2fed82aaf9e5dcefe1a91cdb2e0/pallets/dkg-proposal-handler/src/lib.rs#L480-L488

You will see it will try to `decode_proposal_header` first to get all information needed to actually check if this proposal is on-chain or not, and this information is `chain_id` and the `nonce`.
This is all good and correct, as long as the `decode_proposal_header` return the information, if there is an error (like it fails to decode the proposal header from the data) it will assume this proposal is already there on-chain, which is not the case, and the proposal will be filtered out.

## Root cause

After some debugging, I and @temld4 found the root cause to be that the data gets altered in the DKG Worker here in [`process_unsigned_proposals`](https://github.com/webb-tools/dkg-substrate/blob/1e4e0e53f95cd2fed82aaf9e5dcefe1a91cdb2e0/dkg-gadget/src/worker.rs#L1272):
https://github.com/webb-tools/dkg-substrate/blob/1e4e0e53f95cd2fed82aaf9e5dcefe1a91cdb2e0/dkg-gadget/src/worker.rs#L1308-L1309

the way `pre_signing_proposal_handler` is implemented is that for any `chain_id_type` being `EVM` it will use the hash of the data instead of the data itself, and throw away the original data.
https://github.com/webb-tools/dkg-substrate/blob/1e4e0e53f95cd2fed82aaf9e5dcefe1a91cdb2e0/dkg-gadget/src/worker.rs#L1259-L1270

and now instead of the `decode_proposal_header` try to decode the proposal header (40 bytes) it will try to decode it from the hash (32 bytes) which just fails, and it is impossible to get the required information from just the hash.

## Suggested Solutions

Since the actual data and the header and proposal body are essential for all of the systems that build on top of the DKG, including external systems like the relayers. I would like to suggest instead of doing the hashing here we could include more information inside the `DKGMessage` like how we should process the payload and what style we should use while signing the data.

currently, the `DKGMessage` looks like the following:
```rust
pub struct DKGMessage<AuthorityId> {
	/// Node authority id
	pub id: AuthorityId,
	/// DKG message contents
	pub payload: DKGMsgPayload,
	/// Indentifier for the message
	pub round_id: RoundId,
}
```

we could include a small enum that would define how we should sign the payload.

```rust
pub enum SignatureFormat {
	Default,
	EVM,
    // ... more types on the future?
}
```

That way, other external systems should expect which format depending on the `DKGPayloadKey` stored as the key in the on-chain signed proposals.

This bug was introduced here on this commit: https://github.com/webb-tools/dkg-substrate/commit/1e4e0e53f95cd2fed82aaf9e5dcefe1a91cdb2e0#diff-c38728c577826e5968ed6972f33047c0aa15f569dfea567a04c7e4d8da2c5d30R1259

Related: https://github.com/webb-tools/protocol-solidity/issues/108

shekohex:
This is mainly to address the following issues:

Closes #74 
Closes #146

### DIFF
--- a/dkg-gadget/src/worker.rs
+++ b/dkg-gadget/src/worker.rs
@@ -1171,6 +1171,11 @@ where
 					data: finished_round.payload,
 					signature: finished_round.signature,
 				}),
+			Ok((_chain_id, DKGPayloadKey::RescueTokensProposal(_nonce))) =>
+				Some(ProposalType::RescueTokensSigned {
+					data: finished_round.payload,
+					signature: finished_round.signature,
+				}),
 			Ok((_chain_id, DKGPayloadKey::MaxDepositLimitUpdateProposal(_nonce))) =>
 				Some(ProposalType::MaxDepositLimitUpdateSigned {
 					data: finished_round.payload,
@@ -1303,6 +1308,7 @@ where
 				ProposalType::TokenAdd { data } => data,
 				ProposalType::TokenRemove { data } => data,
 				ProposalType::WrappingFeeUpdate { data } => data,
+				ProposalType::RescueTokens { data } => data,
 				ProposalType::EVMUnsigned { data } => data,
 				ProposalType::MaxDepositLimitUpdate { data } => data,
 				ProposalType::MinWithdrawalLimitUpdate { data } => data,
--- a/dkg-gadget/src/worker.rs
+++ b/dkg-gadget/src/worker.rs
@@ -1171,6 +1171,26 @@ where
 					data: finished_round.payload,
 					signature: finished_round.signature,
 				}),
+			Ok((_chain_id, DKGPayloadKey::MaxDepositLimitUpdateProposal(_nonce))) =>
+				Some(ProposalType::MaxDepositLimitUpdateSigned {
+					data: finished_round.payload,
+					signature: finished_round.signature,
+				}),
+			Ok((_chain_id, DKGPayloadKey::MinWithdrawLimitUpdateProposal(_nonce))) =>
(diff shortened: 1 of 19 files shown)


drewstone (approved):
LGTM

PR #155, opened 2022-02-08
Handle When DKG Signer Changes and the Signed Proposal is Signed by the old DKG.

### LINKED ISSUE #140: [BUG] When DKG Signer get changed will result in an invalid signature for off-chain signed proposals.
Changing the DKG Signer (a la DKG Refresh Proposals) gets changed every new session. and that results in a new DKG Public Key, and if there is a Proposal batch on the off-chain storage got signed by the old signer and when the ocw try to submit these proposals back on-chain, this will fail since now the signer got changed.

Here are some solutions that I have in mind.

1. Trigger a Re-signing again when the DKG Signer changes.
2. ~Instead of just comparing the public key of the signer to the current DKG public key, we could compare it to the old ones too~. (NOT SURE IF THIS IS GOOD OR NOT).
3. Remove batching, and just submit as we go.

shekohex:
Closes #140

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2140,6 +2140,7 @@ dependencies = [
  "frame-support",
  "frame-system",
  "hex",
+ "impl-trait-for-tuples",
  "parity-scale-codec",
  "scale-info",
  "sp-api",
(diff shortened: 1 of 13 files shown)


drewstone:
CI is failing though.

Issue #170, opened 2022-02-16
Organize all non-DKG message broadcasts into a separate file.


drewstone:
# Overview
We currently have a few message types that are not DKG-specific messages that we are gossiping. In an effort to organize the codebase more we should move these out into a separate file. These include:
- PublicKeyBroadcast
- MisbehaviourBroadcast

We should move their gossip handlers into this file.

PR #180, opened 2022-02-21
Organize all non-DKG message broadcasts into a separate file. 

dharjeezy:
Organize all non-DKG message broadcasts into a separate file according to #170


drewstone (changes requested):
Non DKG message is a bad name for readability. Consider creating a separate file for each message and putting inside a directory that is named better. Feel free to brainstorm in slack too.

dharjeezy:
> Non DKG message is a bad name for readability. Consider creating a separate file for each message and putting inside a directory that is named better. Feel free to brainstorm in slack too.

Alright.

drewstone (approved):

Issue #216, opened 2022-03-04
[TASK] Cleaning up, refactor, and better organize repo


dutterbutter:
# Overview

Purpose of this task is to outline a variety of small maintenance chores to improve the repo. 

## Checklist

- [x] #228
- [x] Deduplicate frequently used blocks of code
- [x] Improve the handling of `unwraps()` 
- [ ] #226 
- [x] #227
- [x] #171 
- [x] https://github.com/webb-tools/dkg-substrate/issues/244
- [ ] #245

PR #223, opened 2022-03-08
remove all warnings for dkg-runtime-primitives

dharjeezy:
closes part of #216

### DIFF
--- a/dkg-runtime-primitives/src/handlers/evm/add_token_to_set.rs
+++ b/dkg-runtime-primitives/src/handlers/evm/add_token_to_set.rs
@@ -1,6 +1,6 @@
 use crate::{
 	handlers::{decode_proposals::decode_proposal_header, validate_proposals::ValidationError},
-	ChainIdTrait, ChainIdType, DKGPayloadKey, ProposalHeader,
+	ChainIdTrait, ProposalHeader,
 };
 use codec::alloc::string::ToString;
 use ethereum_types::Address;
--- a/dkg-runtime-primitives/src/handlers/evm/anchor_update.rs
+++ b/dkg-runtime-primitives/src/handlers/evm/anchor_update.rs
@@ -1,6 +1,6 @@
 use crate::{
 	handlers::{decode_proposals::decode_proposal_header, validate_proposals::ValidationError},
-	ChainIdTrait, ChainIdType, DKGPayloadKey, ProposalHeader,
+	ChainIdTrait, ProposalHeader,
 };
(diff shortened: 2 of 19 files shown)


dkg-gadget/src/persistence.rs
```diff
@@ -278,7 +278,7 @@ where
 {
 	let rounds = worker.take_rounds();
 	let next_rounds = worker.take_next_rounds();
-	let time_to_restart = worker.get_time_to_restart(header);
+	let _time_to_restart = worker.get_time_to_restart(header);
```
dutterbutter, line 281: If we are not using the variable `_time_to_restart` can we just have `worker.get_time_to_restart(header);` with no return value? Thoughts?

dkg-gadget/src/messages/misbehaviour_report.rs
```diff
@@ -29,7 +23,7 @@ where
 {
 	// Get authority accounts
 	let header = dkg_worker.latest_header.as_ref().ok_or(DKGError::NoHeader)?;
-	let current_block_number = header.number().clone();
+	let _current_block_number = header.number().clone();
```
dutterbutter, line 26: I think this line can be removed if we are not using it and its only fetching some data rather then setting some data. Thoughts?

dkg-gadget/src/utils.rs
```diff
@@ -31,11 +30,11 @@ pub fn validate_threshold(n: u16, t: u16) -> u16 {
 pub fn set_up_rounds<N: AtLeast32BitUnsigned + Copy>(
 	authority_set: &AuthoritySet<AuthorityId>,
 	public: &AuthorityId,
-	sr25519_public: &sr25519::Public,
+	_sr25519_public: &sr25519::Public,
```
dutterbutter, line 33: Maybe add `TODO:` above fn params we are not currently using?

dkg-gadget/src/persistence.rs
dharjeezy: Alright

dkg-gadget/src/messages/misbehaviour_report.rs
dharjeezy: Alright

dkg-gadget/src/utils.rs
dharjeezy: Alright

drewstone:
Conflicts here from recent merge.

Issue #204, opened 2022-03-02
[TASK] Add validation over EVM addresses


dutterbutter:
**Task summary**

We should be validating EVM addresses within `dkg-runtime-primitives/src/handlers/evm`.

- [ ] Validate in `add_token_to_set.rs`
- [x] Ensure function sig is not zerio in `anchor_update.rs`
- [ ] Validate in `bytes32_update.rs`
- [ ] Validate in `fee_recipient_update.rs`
- [ ] Validate in `fee_update.rs`
- [ ] Validate in `remove_token_from_set.rs`
- [ ] Validate in `rescue_tokens.rs`
- [ ] Validate in `resource_id_update.rs`
- [ ] Validate in `set_treasury_handler.rs`
- [ ] Validate in `set_verifier.rs`

**Specification reference**

- 

**Other information and links**

PR #224, opened 2022-03-08
Ensure function sig is not zero in anchor_update.rs

dharjeezy:
Ensure function sig is not zero in anchor_update.rs according to #204


dkg-runtime-primitives/src/handlers/evm/anchor_update.rs
```diff
@@ -40,3 +44,15 @@ pub fn create<C: ChainIdTrait>(data: &[u8]) -> Result<AnchorUpdateProposal<C>, V
 	// TODO: Ensure function sig is non-zero
```
dutterbutter, line 44: Remove the comment now that we do the check

dkg-runtime-primitives/src/handlers/evm/anchor_update.rs
```diff
@@ -37,6 +41,18 @@ pub fn create<C: ChainIdTrait>(data: &[u8]) -> Result<AnchorUpdateProposal<C>, V
...
+
 	Ok(AnchorUpdateProposal { header, src_chain_id, latest_leaf_index, merkle_root })
 }
+
+fn is_function_signature_valid(data: &[u8]) -> bool {
+	let mut function_sig_bytes = [0u8; 4];
+	function_sig_bytes.copy_from_slice(&data[32..36]);
+	let function_sig = u32::from_be_bytes(function_sig_bytes);
+
+	if function_sig < 0 {
```
drewstone, line 53:
```suggestion
	if function_sig <= 0 {
```

drewstone (changes requested):

Issue #168, opened 2022-02-16
Consolidate public key submission into a single flow.


drewstone:
# Overview
We are currently setting the current public key and next public key in two separate flows: `GENESIS` and not. We should be able to consolidate this into a single flow.

The entry points for this logic is in the `offchain_worker` function of the `pallet-dkg-metadata` and in the `worker.rs`. The basic flows are as follows:

**Genesis flow**
1. The DKG is started and a key is created.
2. The authorities of the DKG gossip this public key and a signature of this public key.
3. The signatures are stored in the offchain worker once enough are gathered.
4. Someone submits enough signatures on-chain and sets the public key using `submit_genesis_public_key_onchain`, which queries the offchain storage and collects the output to submit on-chain.

**Non-genesis flow**
1. The next authorities start a DKG and a key is created. We consider this to be the `next_public_key` or `next_dkg_public_key`.
2. The next authorities of the DKG gossip this public key and a signature of this public key.
3. The signatures are stored in the offchain worker once enough are gathered at a different offchain storage key to the genesis key.
4. Some authority submits enough signatures on-chain and sets the next public key using `submit_next_public_key_onchain`.

As you can see, the flows are nearly identical.

# Task
- [ ] Consolidate this into a single flow

PR #233, opened 2022-03-09
Consolidate public key submission into a single flow

dharjeezy:
Consolidate public key submission into a single flow. #168


dkg-gadget/src/worker.rs
```diff
@@ -860,49 +861,62 @@ where
...
+		log_indicator: &str,
+	) {
+		offchain.set(STORAGE_PREFIX, aggregated_keys, &keys.encode());
+		let submit_at =
+			self.generate_delayed_submit_at(current_block_number.clone(), MAX_SUBMISSION_DELAY);
+		if let Some(submit_at) = submit_at {
+			offchain.set(STORAGE_PREFIX, submit_keys, &submit_at.encode());
+		}
+
+		trace!(
```
drewstone, line 910: Let's remove this log, it doesn't look correct.

dharjeezy: Alright

PR #247, opened 2022-03-15
Add docs to pallets

drewstone:


pallets/dkg-metadata/src/lib.rs
```diff
@@ -12,6 +12,85 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! # DKG Metadata Module
+//!
+//! A pallet to manage metadata about the DKG and DKG proposal system
```
dutterbutter, line 17: nit: no period at end

dutterbutter (approved):

PR #250, opened 2022-03-15
updated benchmarks

salman01zp:
**Summary of changes**
- updated benchmark of dkg-proposals pallet
- updated benchmark of dkg-proposal-handler pallet
-  generated weight.rs file for dkg-proposal pallet

Closes


runtime/Cargo.toml
```diff
@@ -60,6 +60,8 @@ pallet-transaction-payment = { git = "https://github.com/paritytech/substrate",
 pallet-transaction-payment-rpc-runtime-api = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.17", default-features = false }
 pallet-authorship = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.17", default-features = false }
 pallet-session = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.17", default-features = false }
+pallet-society = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.17", default-features = false }
```
drewstone, line 63: Do we need this dependency?

runtime/Cargo.toml
```diff
@@ -60,6 +60,8 @@ pallet-transaction-payment = { git = "https://github.com/paritytech/substrate",
...
+pallet-collective = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.17", default-features = false }
```
drewstone, line 64: Do we need this dependency?

pallets/dkg-proposal-handler/src/benchmarking.rs
```diff
@@ -60,19 +60,19 @@ benchmarks! {
 		let n in 0..(T::MaxSubmissionsPerBatch::get()).into();
 		let dkg_pub_key = ecdsa_generate(KEY_TYPE, None);
 		DKGPallet::<T>::set_dkg_public_key(dkg_pub_key.encode());
-
+		let caller: T::AccountId = whitelisted_caller();
 		let mut signed_proposals = Vec::new();
 		for i in 0..n as usize {
 			let tx = TransactionV2::EIP2930(mock_eth_tx_eip2930(i as u8));
-			let proposal = <Pallet<T> as ProposalHandlerTrait>::force_submit_unsigned_proposal(Proposal::Unsigned {
+			let proposal = Pallet::<T>::force_submit_unsigned_proposal(RawOrigin::Signed(caller.clone()).into(),Proposal::Unsigned {
```
drewstone, line 67: We need to ensure that this works, I believe force submit should be called from a Root origin.

drewstone (changes requested):
Some preliminary questions.

pallets/dkg-proposal-handler/src/benchmarking.rs
salman01zp: yes it required Root Origin, updated

runtime/Cargo.toml
salman01zp: No, we don't require this dependency but  I am not able to build node without adding them.  I have asked the same  question in  [Issue](https://github.com/paritytech/substrate/issues/10548 )

salman01zp: Removed those dependencies. runtime-benchmarks feature flag was needed to be enabled for polkadot-cli  in ./node/cargo.toml

for details regarding the issue :  [Issue Link](https://substrate.stackexchange.com/questions/240/missing-successful-origin-in-implementation-when-running-runtime-benchmarks)

runtime/Cargo.toml
salman01zp: No, removed that dependency.

pallets/dkg-proposal-handler/Cargo.toml
```diff
@@ -15,7 +15,7 @@ targets = ["x86_64-unknown-linux-gnu"]
 codec = { package = "parity-scale-codec", version = "2.0.0", features = ["derive"], default-features = false }
 scale-info = { version = "1.0", default-features = false, features = ["derive"] }
 
-
+pallet-dkg-proposals = { path = "../dkg-proposals", default-features = false }
```
drewstone, line 18: I don't believe this should be needed in the main dependencies. If anything, it should be an optional dependency `optional = true` and you should be able to remove it from `dev-dependencies` as well.

drewstone (changes requested):
Looks good, can we update the title to assert this is ready and not WIP?

pallets/dkg-proposal-handler/Cargo.toml
salman01zp: removed that dependency and runtime-benchmark features that were used for benchmarking.

drewstone (approved):

PR #261, opened 2022-03-22
Akilesh/proposer set tests

saiakilesh:
**Summary of changes**
- 

Closes


drewstone (changes requested):
Looks good. Spacing is off on the tests and should be fixed before merge. 

Can the proposers vote on the governor again and again too out of curiosity?

PR #264, opened 2022-03-23
Change spacing to 2 spaced tabs

drewstone:


dutterbutter:
@drewstone we should add `prettier` and a config file for making future formatting within these files easier.

PR #281, opened 2022-04-04
Fix persistent issues and add force rotate

drewstone:


dkg-gadget/src/persistence.rs
```diff
@@ -100,6 +100,43 @@ where
 	}
 }
 
+pub(crate) fn load_stored_key<B, C, BE>(
```
dharjeezy, line 103: You can help add some comments on here.

dharjeezy:
LGTM!
Just to make CI green is what is remaining.

PR #302, opened 2022-05-02
Convert serial/synchronous rounds to parallel/asynchronous rounds (#301)

drewstone:


tbraun96:
Linting with `clippy` when using the dependency ``socket2`` causes errors. Evidently, Clippy internally sets an invalid build target, and as such, we get the errors that we do in the pipeline. For a potential workaround, [we can tell clippy to only check the workspace](https://github.com/rust-lang/cargo/issues/8143). This will require a change to the github actions pipeline on a single line.

Linting with clippy works when we use: ``cargo clippy -Zunstable-options -- -D warnings``

tbraun96:
The reason why `tests` is failing the pipeline is because we are using the rust toolchain `wasm32-unknown-unknown`, yet, when compiling `socket2`, we get an error because socket2 does not (yet) support wasm.

dutterbutter:
@tbraun96 hmmmm it doesn't look like too much momentum in having socket2 support wasm based on this [issue](https://github.com/rust-lang/socket2/issues/268). It mentions perhaps disabling tokio `net` feature might work, I don't know if that is a great idea tho.

tbraun96:
> @tbraun96 hmmmm it doesn't look like too much momentum in having socket2 support wasm based on this [issue](https://github.com/rust-lang/socket2/issues/268). It mentions perhaps disabling tokio `net` feature might work, I don't know if that is a great idea tho.

The issue is that WASM cannot access File I/O, and as such, it cannot access the underlying file descriptors required by async programming. Custom socket opts are requires for libp2p since NAT traversal, depending on the technique, requires `SO_REUSE` set.

drewstone:
@dutterbutter can you fix the link and license CI errors on this branch?

dutterbutter:
@drewstone yes no problem

PR #312, opened 2022-06-03
[feat] Remove expired unsigned proposals from queue

### LINKED ISSUE #307: [TASK] Handle cleaning of old unsigned proposals - add lifetimes to UnsignedProposal
**Issue summary**
In a worst-case scenario, unsigned proposals fail to get signed and clog the unsigned proposal queue. We should have a way of attaching temporal (time-based) data to unsigned proposals when they are added and cleaning these proposals from the storage when some time has elapsed. At least in this case, we can keep state growth down to a minimum in the event the DKG fails to sign things over time.

**Ideas**
The way we would do this is by adding lifetimes to each UnsignedProposal that reaches the `pallet-proposal-handler`. Then we would check every so often in the `Hooks` if there are unsigned proposals and check there lifetimes, removing if they've become outdated.

1xstj:
**Summary of changes**
- Add timestamp to UnsignedProposals in UnsignedProposalQueue
- Introduce `UnsignedProposalExpiry` config to set expiry time for unsigned proposals
- Remove expired unsigned proposals in `on_idle` hook

Closes #307

### DIFF
--- a/dkg-runtime-primitives/src/proposal.rs
+++ b/dkg-runtime-primitives/src/proposal.rs
@@ -240,3 +240,13 @@ pub trait ProposalHandlerTrait {
 }
 
 impl ProposalHandlerTrait for () {}
+
+/// An unsigned proposal represented in pallet storage
+/// We store the creation timestamp to purge expired proposals
+#[derive(Debug, Encode, Decode, Clone, Eq, PartialEq, scale_info::TypeInfo)]
+pub struct StoredUnsignedProposal<Timestamp> {
+	/// Proposal data
+	pub proposal : Proposal,
+	/// Creation timestamp
+	pub timestamp : Timestamp
+}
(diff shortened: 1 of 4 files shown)


dkg-runtime-primitives/src/proposal.rs
```diff
@@ -240,3 +240,13 @@ pub trait ProposalHandlerTrait {
 }
 
 impl ProposalHandlerTrait for () {}
+
+/// An unsigned proposal represented in pallet storage
+/// We store the creation timestamp to purge expired proposals
+#[derive(Debug, Encode, Decode, Clone, Eq, PartialEq, scale_info::TypeInfo)]
+pub struct StoredUnsignedProposal<Timestamp> {
+	/// Proposal data
+	pub proposal : Proposal,
```
drewstone, line 249: Extraneous spaces.
```suggestion
	pub proposal: Proposal,
```

dkg-runtime-primitives/src/proposal.rs
```diff
@@ -240,3 +240,13 @@ pub trait ProposalHandlerTrait {
...
+	/// Creation timestamp
+	pub timestamp : Timestamp
```
drewstone, line 251:
```suggestion
	pub timestamp: Timestamp
```

drewstone (changes requested):
Looks great. Does this have any effects to the runtime API that will be consumed by the offchain DKG protocol? We will need to double-check the runtimes to ensure that we return the UnsignedProposal as is expected by the worker and not this new struct (or update it accordingly).

Lastly, we should ensure the E2E tests work locally at the very least before merging this change.

PR #316, opened 2022-06-07
Minimal P2P Reputation

### LINKED ISSUE #309: [SPEC] Handling peer to peer reputation at gossip-engine level
## Overview

We need a solution for handling p2p reputation at the gossip-engine level when we receive good and bad messages.

shekohex:
**Summary of changes**
**Added a p2p reputation in the following cases**:

- [x] When a message can’t be decoded/is invalid.
- [x] When a message has already been seen more than N times (to prevent network flooding)
- [ ] When the message signature is invalid.
- [x] On the other hand, we add a good reputation when we receive a message Not being seen by us, or for the first time.

Closes #309

### DIFF
--- a/dkg-gadget/src/meta_async_rounds/dkg_gossip_engine/network.rs
+++ b/dkg-gadget/src/meta_async_rounds/dkg_gossip_engine/network.rs
@@ -38,18 +38,21 @@
 //! peers or only to a specific peer. on the other end, the DKG message is received by the DKG
 //! engine, and it is verified then it will be added to the Engine's internal stream of DKG
 //! messages, later the DKG Gadget will read this stream and process the DKG message.
-use crate::metrics::Metrics;
+use crate::{
+	meta_async_rounds::dkg_gossip_engine::ReceiveTimestamp, metrics::Metrics, worker::LatestHeader,
+};
 use codec::{Decode, Encode};
 use dkg_primitives::types::{DKGError, SignedDKGMessage};
 use dkg_runtime_primitives::crypto::AuthorityId;
 use futures::{FutureExt, Stream, StreamExt};
 use linked_hash_set::LinkedHashSet;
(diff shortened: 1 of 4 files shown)


dkg-gadget/src/meta_async_rounds/dkg_gossip_engine/network.rs
```diff
@@ -154,19 +153,12 @@ const MAX_PENDING_MESSAGES: usize = 8192;
...
-	/// Reputation change when a peer sends us any message that is not invalid.
-	pub const ANY_MESSAGE_REFUND: Rep = Rep::new(1 << 4, "Any message (refund)");
 	/// Reputation change when a peer sends us a message that we didn't know about.
 	pub const GOOD_MESSAGE: Rep = Rep::new(1 << 7, "Good message");
-	/// Reputation change when a peer sends us a bad message.
-	pub const BAD_MESSAGE: Rep = Rep::new(-(1 << 12), "Bad message");
 	/// We received an unexpected message packet.
 	pub const UNEXPECTED_MESSAGE: Rep = Rep::new_fatal("Unexpected message packet");
+	/// Reputation change when a peer sends us the same message over and over.
+	pub const DUPLICATE_MESSAGE: Rep = Rep::new(-(1 << 12), "Duplicate message");
```
drewstone, line 161: The only issue here is that in certain parts of the protocol, we might intentionally send messages more than once.

shekohex: From the same peer? like the same peer sending us the same exact message (with the same hash) more than one time?

Also, in that case, if you can give me an example of when we will use that, we can add a simple atomic counter to keep track of how many times this peer sent us the exact same message, and if we reached `n` times we say ok this too much and maybe they are flooding the network so we start to add neg rep to them.

PR #317, opened 2022-06-08
[fix] Update failing benchmarks of dkg-metadata

1xstj:
**Summary of changes**
- Fix benchmarks failure in dkg-metadata
```
failures:
---- benchmarking::benchmark_tests::test_benchmarks stdout ----
failing benchmark tests:
submit_public_key: "InvalidPublicKeys"
submit_next_public_key: "InvalidPublicKeys"
submit_misbehaviour_reports: "InvalidMisbehaviourReports"
```

Closes


pallets/dkg-metadata/src/benchmarking.rs
```diff
@@ -28,7 +28,7 @@ use dkg_runtime_primitives::{
 use frame_benchmarking::{benchmarks, impl_benchmark_test_suite};
 use frame_system::RawOrigin;
 use sp_io::crypto::{ecdsa_generate, ecdsa_sign_prehashed, sr25519_generate, sr25519_sign};
-use sp_runtime::{key_types::AURA, traits::TrailingZeroInput, Permill};
+use sp_runtime::{key_types::{AURA, DKG}, traits::TrailingZeroInput, Permill};
```
drewstone, line 31: Just in case it got lost in discussion, there is no `key_type::DKG`. The key_type is named `KEY_TYPE` in the `dkg_runtime_primitives` so that is where you'll want to import it.

pallets/dkg-metadata/src/benchmarking.rs
```diff
@@ -48,8 +48,8 @@ fn mock_signature(pub_key: sr25519::Public, dkg_key: ecdsa::Public) -> (Vec<u8>,
 	(msg, signature.encode())
 }
 
-fn mock_pub_key() -> sr25519::Public {
-	sr25519_generate(AURA, None)
+fn mock_pub_key() -> ecdsa::Public {
+	ecdsa_generate(ECDSA, None)
```
drewstone, line 52: It will look like the following once the proper import is given.
```suggestion
	ecdsa_generate(KEY_TYPE, None)
```

drewstone (changes requested):

drewstone:
Can you include the new weights if the benchmarks work now? When we fix benchmarks we should always re run the weights and update. What do you think @1xstj?

drewstone (approved):

PR #320, opened 2022-06-11
Multiple sign protocol execution

drewstone:
**Summary of changes**
Attempts to run the signing protocol for other subsets of signers to cover the cases that other signers misbehave and stop participating. The rudimentary idea is to randomly select other sets of $t+1$ signers and to simultaneously execute $((t+1 / 2) + 1$ signing protocols in parallel.

Nonetheless, currently it seems like they don't execute in parallel and only one continues to execute. CC @tbraun96 if you have ideas here.

Closes


drewstone:
Hitting sign errors from within `multi-party-ecdsa`, likely the multiple executions are intertwining with one another and causing breaking errors. Need to figure out a way to ensure they execute in parallel without processing one another's messages.

PR #330, opened 2022-06-17
Fix signer set bug

drewstone:


drewstone:
I think this fixes it @shekohex. Now we should test adding Charlie back and increasing thresholds to continue to fish out more bugs.

shekohex:
> Now we should test adding Charlie back and increasing thresholds to continue to fish out more bugs. 

Yup, Will be working on that.

PR #331, opened 2022-06-20
Update Threshold on unjail

shekohex:


shekohex:
Updating threshold will be through voting (democracy pallet in our case).

No need for this PR then.

PR #332, opened 2022-06-20
Fix Keygen Threshold updates

shekohex:
### Current Behaviour: 

Start all three nodes (Alice, Bob, Charlie) at block eight before they rotate, and after the keygen, stop/kill the Charlie node. The other nodes will notice that and start reporting Charlie in the next keygen; then Charlie gets jailed. The Keygen threshold gets updated automatically to adjust for the new changes, and a new Keygen (forced) kicks in; later on, Charlie’s node starts running again, and since he is currently jailed, he does not participate in the Keygen.

Next, Charlie gets unjailed, and nothing changed in the system itself; only Alice and bob, in this case, are the best authorities. Fast forward, a new Call to change the Keygen threshold (using sudo in my case, but it would be through Democracy pallet) to be adjusted to 3 from 2; we notice now that the threshold gets updated, the next best authorities now includes Charlie, However, for some reason, we start to see invalid signature during the next refresh, so this part is kinda broken, meaning the best authorities is not going to be updated to be the same as the next best authorities. **However, I tried to do another sudo call to force change the current best authorities to be the next best authorities**, that’s when everything started to get back to normal, and the Keygen/Signing was working as usual.

### What we expect:

The updates should work without the need for the sudo call to force change the current best authorities to be the next best authorities.

### What is fixed now

Alright, So far so good, I’ve tested the following Flows:
1. Start with 3 Nodes, Keygen Threshold is 2, the first Keygen runs without any issues, now upgrade the Keygen Threshold to 3, wait till next session, you should see now that the 3 nodes are running, and it continues to rotate each session ✅ 

2. While on the same flow as #No.1 now try to downgrade the Keygen Threshold (manually) and you should notice the next session that it only picked 2 nodes, and it should continue to run and rotates. ✅ 

3. Last but not least, try to Upgrade the Keygen Threshold again to 3 and wait for the next session, and then kill one of the 3 nodes, you will notice that the other two nodes will report it then it will get jailed, next is that the keygen threshold will drop to 2 (instead of 3) and a new Keygen will happen immediately.

Now after another session try to start the killed node again, unjail it, then upgrade the Keygen threshold to 3, you should see the three nodes running together again, and they continue to rotate. ✅


drewstone (commented):
Looks like good fixes so far!

PR #343, opened 2022-07-28
Add signed handler

drewstone:
**Summary of changes**
- Adds the on signed proposal handler that can call into other pallets that implement it.
- Updates webb proposals versions and moves some of proposal utilities into the `webb-proposals` repo.


1xstj (approved):
LGTM

PR #359, opened 2022-08-30
[TASK] Properly handle errors from mp-ecdsa library (parse bad_actors)

### LINKED ISSUE #357: [TASK] Properly handle errors from mp-ecdsa library (parse bad_actors)
**Issue summary**
We aren't parsing `bad_actors` from a lot of errors returned from the keygen/offline/voting stages of the underlying DKG protocols. We instead are returning either `DKGError::CriticalError`s or `DKGError::GenericError`s and turning the actual errors into strings.

Instead, we should be match on the potential error to see if it is one that contains `bad_actors`. Namely, we want to parse out errors from the `multi-party-ecdsa` library that are of `ErrorType` type:

The `ErrorType` contains among other things `bad_actors` which are the indices of the parties in the DKG protocol. Since we are instantiating the participants of the protocol using the `best_authorities`, we must map them back into the `best_authorities` rather than the entire validator set or authorities running the chain.
```
pub struct ErrorType {
    pub error_type: String,
    pub bad_actors: Vec<usize>,
    pub data: Vec<u8>,
}
```

For any nuance, we effectively need to map the reported index back into the `best_authorities` set and then report that Authority.

## Nuance / Questions
- The `bad_actors` of errors during the Keygen process are definitely indexed as they are in the best_authorities set.
- It is not exactly clear to me right now if the `bad_actors` of errors during the Offline/Sign processes are indexed by the best_authorities or the chosen signer subset of that protocol execution.

1xstj:
**Summary of changes**
- Unwrap mp-ecdsa errors to expose bad_actors returned from lib
- Refactor handle_dkg_error fn to index by best_authorities
- Some logging target improvements

Closes #357

### DIFF
--- a/dkg-primitives/src/types.rs
+++ b/dkg-primitives/src/types.rs
@@ -269,8 +269,8 @@ impl fmt::Display for DKGError {
 			CriticalError { reason } => format!("Critical error: {}", reason),
 			GenericError { reason } => format!("Generic error: {}", reason),
 			StartOffline { reason } => format!("Unable to start Offline Signing: {}", reason),
-			NoAuthorityAccounts => format!("No Authority accounts found!"),
-			NoHeader => format!("No Header!"),
+			NoAuthorityAccounts => "No Authority accounts found!".to_string(),
+			NoHeader => "No Header!".to_string(),
 			SignMisbehaviour { reason, bad_actors } =>
 				format!("SignMisbehaviour : reason: {:?},  bad actors: {:?}", reason, bad_actors),
 		};
--- a/dkg-primitives/src/types.rs
+++ b/dkg-primitives/src/types.rs
@@ -259,6 +259,7 @@ pub enum DKGError {
 	NotListeningForPublicKey,
 	NoAuthorityAccounts,
 	NoHeader,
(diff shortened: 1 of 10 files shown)


drewstone (commented):
LG so far, I'm curious your thoughts on how we can ensure we are mapping bad actors from signing protocol into the authority set properly. Have you been able to locate which indices they are exactly?

EDIT: I think we will still receive the indexed list of signers in the best authorities set so we can just grab their indices 👍

dkg-primitives/src/types.rs
```diff
@@ -259,6 +259,7 @@ pub enum DKGError {
 	NotListeningForPublicKey,
 	NoAuthorityAccounts,
 	NoHeader,
+	MPCSignError { reason: String, bad_actors: Vec<usize> },
```
drewstone, line 262: If you're adding MPCSignError and using it instead of `SignMisbehaviour`, can you do the same with the Keygen error too and remove their counterparts too?

drewstone (changes requested):

drewstone (approved):
LGTM

PR #365, opened 2022-08-31
Split gossip protocol types between keygen/signing

### LINKED ISSUE #363: [TASK] Consider splitting gossip protocol types between keygen/signing
**Issue summary**
We currently process all incoming DKG messages within the same stream and flow, which could pose issues if there are ever issues on the networking layer. Instead, we can consider separating between keygen messages and signing messages. We would then have 2 streams for each of the protocols.

We would need to:
1. Create different protocol names for the keygen and sign sub-protocols.
2. Share the same networking code.
3. Handle incoming messages separately in new handlers `process_incoming_keygen_msg` and `process_incoming_sign_msg`.
4. Gossip outgoing messages separately using protocol specific `keygen_gossip_handler` and `sign_gossip_handler`.

- https://github.com/paritytech/substrate/blob/master/client/network/src/transactions.rs

1xstj:
**Summary of changes**

- Refactor gossip engine and worker to support multiple network streams based on `protocol_name`
- Create seperate gossip engines for keygen/signing

Closes #363


drewstone (approved):
LGTM

PR #381, opened 2022-09-08
Rework the `DKGWorker`.

### LINKED ISSUE #375: [SPEC] Rework `DKGWorker` to be more share-able across different async tasks
## Overview 

In the context of [`DKGWorker::run`](https://github.com/webb-tools/dkg-substrate/blob/bbc5a9071d7962f59200f76d090a185d31b84b39/dkg-gadget/src/worker.rs#L1313) method we _start_ 4 Tasks to run and watch over these results, but right now we are using [`tokio::select!`](https://docs.rs/tokio/1.21.0/tokio/macro.select.html) macro, and as per its docs it outlines the following [**"running all async expressions on the current task, the expressions are able to run concurrently but not in parallel."**](https://docs.rs/tokio/1.21.0/tokio/macro.select.html#runtime-characteristics) and for us this means for the current flow for example if `finality_notifications` future/stream finished first and at the same time we have a `dkg_msg` from the `gossip_engine` stream, the select macro could pick the block notification and completely ignoring "discarding" the message from the dkg stream resulting in a completely loss of the information here.

## Research

We should instead use `tokio::spawn` for each of these tasks and run them independently from each others in parallel, this from also the above docs it mentions the following: "If parallelism is required, spawn each async expression using [tokio::spawn](https://docs.rs/tokio/1.21.0/tokio/task/fn.spawn.html) and pass the join handle to select!"

## Examples 

Here is a good example of what I do reassemble in my head of a good solution for the current issue:

```rust
pub(crate) async fn run(mut self) {
	// ... snip ...
	let finality_notifications_task = async move {
		// ...
	};
	let import_notifications_task = async move {
		// ...
	};
	let misbehaviour_task = async move {
		// ...
	};
	let error_handler_task = async move {
		// ...
	};
	let dkg_messages_task = async move {
		// ...
	};

	let handles = vec![
		tokio::spawn(finality_notifications_task),
		tokio::spawn(import_notifications_task),
		tokio::spawn(misbehaviour_task),
		tokio::spwan(error_handler_task),
		tokio::spawn(dkg_messages_task),
	];
	let _ = futures::future::select_all(handles).await;

	// ... snip ...
}

```

## Questions/Issues

- [x] Current Implementation of the `DKGWorker` is not shareable across different tasks, e.g. most functions/methods takes `&mut self` which is not feasible to share with a mutable state, we shall use async locks or other things to make it possible.
- [x] decouple the worker into smaller states when possible, so that it could be easily shareable between different tasks.

shekohex:
**Summary of changes**
- See #375 

Closes #375


dkg-gadget/src/gossip_engine/network.rs
```diff
@@ -328,43 +349,53 @@ impl<B: Block + 'static> GossipHandler<B> {
...
+					_ => {},
+				}
 			}
-		}
+		});
+
+		// a timer that fires every few ms to check if there are messages in the queue, and if so,
+		// notify the listener.
+		let self1 = self.clone();
+		let mut timer = tokio::time::interval(core::time::Duration::from_millis(100));
```
drewstone, line 371: What are the downsides of firing so often? Are there any?

dkg-gadget/src/gossip_engine/network.rs
```diff
@@ -328,43 +349,53 @@ impl<B: Block + 'static> GossipHandler<B> {
...
+		});
+
+		let network_events_task = tokio::spawn(async move {
+			while let Some(event) = event_stream.next().await {
+				self.handle_network_event(event).await;
+			}
+		});
+
+		// wait for the first task to finish or error out.
+		let _result = futures::future::select_all(vec![
```
drewstone, line 389: Can we add more documentation on why we wait for just one of them to complete?

dkg-gadget/src/gossip_engine/network.rs
```diff
@@ -328,43 +349,53 @@ impl<B: Block + 'static> GossipHandler<B> {
...
-					// check if we have any messages in our queue.
-					let read_lock = self.message_queue.read();
-					if !read_lock.is_empty() {
-						// notify the caller that we have messages in the queue.
-						self.message_notifications_channel.send(()).unwrap();
-					}
-				},
+
+		let self0 = self.clone();
+		let incoming_messages_task = tokio::spawn(async move {
```
drewstone, line 357: Would be good to add more documentation here

dkg-gadget/src/worker.rs
```diff
@@ -1386,84 +1427,96 @@ where
...
-					if let Some(error) = error {
-						log::debug!(target: "dkg_gadget::worker", "Going to handle Error");
-						self.handle_dkg_error(error);
-					}
-				},
+			.filter_map(move |_| futures::future::ready(signing_gossip_engine.peek_last_message()));
+
+		// create a task for each of the following tasks:
+		// 1. handle import notifications
+		// 2. handle finality notifications
```
drewstone, line 1444: Can we remove the finality notifications now actually? They're not necessary

drewstone: Fewer tasks the better, especially for ones we know we don't need.

drewstone (commented):
LG, would consider all the ways we can improve the readability and documentation.

dkg-gadget/src/worker.rs
```diff
@@ -1386,84 +1427,96 @@ where
...
+		// 6. handle dkg errors.
+		let mut self1 = self.clone();
+		let import_notifications_task = tokio::spawn(async move {
+			while let Some(notification) = self1.import_notifications.next().await {
+				log::debug!(target: "dkg_gadget::worker", "Going to handle Import notification");
+				self1.handle_import_notification(notification);
 			}
-		}
+		});
+		let mut self2 = self.clone();
```
drewstone, line 1456: Can these be broken out into helper functions at all?

dkg-gadget/src/gossip_engine/network.rs
shekohex: Not at all, this task running on its own, there is no blocking nor it should affect any other tasks that are running along side it, even more this task could be running on totally separate thread, the runtime decides that for us.

drewstone (commented):
Thoughts on breaking the spawned threads in the run method into helper functions? Is it possible?

shekohex:
> Thoughts on breaking the spawned threads in the run method into helper functions? Is it possible?

Yes, we can do that in a smaller functions/methods.

dkg-gadget/src/worker.rs
shekohex: I will try to remove it and see if things are working normally.

dkg-gadget/src/gossip_engine/network.rs
shekohex: done.

drewstone (approved):
LGTM

PR #383, opened 2022-09-10
Fix misbehaviour updating

drewstone:


drewstone:
@nepoche can you check that I added the DVC solidity-fixtures here correctly? Seemed like it was the only thing remaining preventing success maybe.

PR #387, opened 2022-09-21
pallet-light-proposals: Add logic for creating proposals based on log/storage proofs (EVM)

saiakilesh:
**Summary of changes**
- 

Closes


pallets/light-proposals/eth-types/Cargo.toml
```diff
@@ -0,0 +1,49 @@
+[package]
```
drewstone, line 1: @saiakilesh in the future, please try to avoid cloning packages that can simply be imported. We have this package written in the `pallet-eth2-light-client` so I don't see the reason why this is being added here. I will try to clean up this PR a bit this week so don't worry about removing it (I will). This isn't good practice so please avoid it from here on.

drewstone: Unless I'm missing some reason why we're adding this back here?

PR #395, opened 2022-09-26
[SPEC] Make DKG Manages when a Session should end

### LINKED ISSUE #394: [SPEC] Make DKG Manages when a Session should end
## Current Logic

The current signing logic in dkg gadget listens to new block imports and looks for a change in authorities. When a change in authorities is detected it creates a new signing key and uploads onchain for the NextDkgPublicKey

## Proposed Logic

The dkg gadget has a constant time in which it would sign new keys (lets say 90% of session period in runtime), this should be equal/less than the session period set in the runtime. Every interval of this constant time the dkg gadget would execute keygen and upload signed key on-chain. The succesful upload of the new public key is required for a successful rotation (we should check in `should_end_session` if the nextPublicKey and nextPublicKeySignature are stored onchain)

```rs
fn should_end_session(now: BlockNumber) -> bool {
	// 1. ensure the period has passed
	let offset = Offset::get();
	let period_passed = now >= offset && ((now - offset) % Period::get()).is_zero()
	// 2. ensure nextPublicKey exists
	let is_next_public_key = nextDkgPublicKey.is_some();
	// 3. ensure nextPubkeysign exists
	let is_next_public_key_sign = nextDkgPublicKeySign.is_some();
	period_passed && is_next_public_key && is_next_public_key_sign
}
```

### Proposed Changes

- [ ]  Refactor dkg gadget to create new signing keys right before session ends
- [ ]  Refactor session change logic to depend on upload of session keys
- [ ]  Add retry logic for keygen and signing

### Questions

- This new logic could lead to uneven session periods, we assume here that this is non-critical
- If a signedKey upload fails, the retry again will happen only after the fixed interval, maybe we need a error handler retry?
    - We should retry the keygen creation process for n times and if it still fails we report misbehaviour
    - We should retry the signing the already generated keygen for n times and if it still fails we report misbehaviour

1xstj:
**Summary of changes**
 
- [X] Refactor dkg gadget to create new signing keys right before session ends (based on session progress threshold)
- [X]  Refactor session change logic to depend on upload of NextDkgPublicKey and NextDkgPublicKeySignature

Closes #394

### DIFF
--- a/dkg-gadget/src/async_protocols/blockchain_interface.rs
+++ b/dkg-gadget/src/async_protocols/blockchain_interface.rs
@@ -146,6 +146,7 @@ where
 	) -> Result<(), DKGError> {
 		// Call worker.rs: handle_finished_round -> Proposal
 		// aggregate Proposal into Vec<Proposal>
+		log::info!(target: "dkg", "PROCESS VOTE RESULT : round_id {:?}, signature : {:?}", round_id, signature);
 		let payload_key = unsigned_proposal.key;
 		let signature = convert_signature(&signature).ok_or_else(|| DKGError::CriticalError {
 			reason: "Unable to serialize signature".to_string(),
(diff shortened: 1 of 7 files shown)


shekohex (changes requested):

drewstone (approved):
LGTM

PR #414, opened 2022-10-12
Keygen protocol Improvements (retry, edge case, ..etc)

### LINKED ISSUE #411: [SPEC] Retry the Keygen Protocol `N` Times before considering it a misbehaviour
# Overview 

When we reach the point that the Keygen stall, we start the misbehaviour reporting, this is useless in the case we are running `1-out-of-2` DKG, since we need _`t+1`_ reports to consider this misbehaviour. 

In other case where the DKG is running `t-of-n` and `n > t + 1` and the Keygen Stall, we should not immediately start the misbehaviour report, we should retry again.

## Research

* **Case 1**: We are running `1-of-2` DKG Protocol, if the Keygen failed, we must retry again until it work, the misbehaviour reporting is useless in this case.

* **Case 2**:  We are running `t-of-n` DKG Protocol, where `n > t + 1`, and the Keygen failed, we should update our state that the Keygen stalled and retry again, in the next finality notification up till `N` times or the Keygen Protocol Complete. if we still failing after `N` we consider it a misbehaviour case, and we start reporting the misbehaved parties.

In my opinion, Case 1 could be considered as a _sub_ case of Case 2, where Case 2 is the more general case, and Case 1 is special case where we will ignore the max retires count and just keep retrying the keygen protocol over and over.

## Examples 

Consider `MAX_KEYGEN_RETIRES` to be `5`.

# Questions/Issues

1. **When we should restart the Keygen Protocol?**
A. Since all nodes are working on the finality notification, they will detect that the Keygen has been stalled at the same time, each node should set `keygen_stalled` to `true` and each finality block notification we check for the following:
	a. Is there any Keygen in progress? is it stalled? and we have `keygen_stalled` set to `true`?
	b. How many times we retried the Keygen protocol? and if it is over the `MAX_KEYGEN_RETIRES` or not?
after this, depending on the answers, we see if we shall either start a new Keygen protocol again, or report the misbehaviours. 

## Diagrams

image

shekohex:
**Summary of changes**
- Spec #411 

Closes #411


dkg-gadget/src/worker.rs
```diff
@@ -923,15 +932,39 @@ where
...
-				if rounds.keygen_has_stalled(*header.number()) {
-					debug!(target: "dkg_gadget::worker", "🕸️  QUEUED DKG STALLED: round {:?}", queued.id);
+				let keygen_stalled = rounds.keygen_has_stalled(*header.number());
+				let (current_attmp, max, should_retry) = {
+					// check how many authorities are in the next best authorities
+					// and then check the signature threshold `t`, if `t+1` is greater than the
+					// number of authorities and we still have not reached the maximum number of
+					// retries, we should retry the keygen
+					let n = next_best.len();
+					let t = self.get_signature_threshold(header) as usize;
```
drewstone, line 942: If this is next keygen then you need to use next signature threshold.

dkg-gadget/src/worker.rs
```diff
@@ -923,15 +932,39 @@ where
...
+					// in this case, if t + 1 is less than or equal n, we should retry the keygen
+					// indefinitely.
+					let max_retries = if t + 1 > n { MAX_KEYGEN_RETRIES } else { 0 };
```
drewstone, line 945: I don't understand this, why would this ever happen? On chain this is never able to happen.

drewstone (changes requested):
LG but need some small fixes I think.

dkg-gadget/src/worker.rs
shekohex: it should be the current `t` since we will create a misbehaviour reports on the current session/round. and Since we do not want the Misbehaviour reports to be useless, which would be the case if the number of reports (which should be equal to `t+1`) is less than `t+1`. But I will modify that, it should only be if `n` equal to `t+1` that means if we did try to misbehaviour report someone it will fail. Example: n = 2 and t = 1, we need at least `t+1 = 2` reports to report one of these nodes, and since it will only be one report.

dkg-gadget/src/worker.rs
shekohex: Since we do not want the Misbehaviour reports to be useless, which would be the case if the number of reports (which should be equal to `t+1`) is less than `t+1`. But I will modify that, it should only be if `n` equal to `t+1` that means if we did try to misbehaviour report someone it will fail. Example: n = 2 and t = 1, we need at least `t+1 = 2` reports to report one of these nodes, and since it will only be one report.

dkg-gadget/src/worker.rs
shekohex: my bad, it should be the next signature threshold, I found that in the DKG pallet here: https://github.com/webb-tools/dkg-substrate/blob/826d389aa57f67a3f87ef790b021e714e309a815/pallets/dkg-metadata/src/lib.rs#L911-L916

drewstone:
My understanding is that the way this should work is:
1. We retry keygen some # of times (treating $X$-of-2 for $X \in {1, 2}$ as a special case).
2. Once that retry counter has exhausted we report misbehaviours
3. Once a successful report gets on-chain we jail the party and the next best authority set should update.
4. Once we detect this, we retry keygen with the new next best authorities and repeat (1)

shekohex:
> My understanding is that the way this should work is:
> 
> 1. We retry keygen some # of times (treating X-of-2 for X∈1,2 as a special case).
> 2. Once that retry counter has exhausted we report misbehaviours
> 3. Once a successful report gets on-chain we jail the party and the next best authority set should update.
> 4. Once we detect this, we retry keygen with the new next best authorities and repeat (1)
> 
> Let me know if you don't think this is the way we should go @shekohex @1xstj or if we all agree we should ensure the implementation matches this.

Exactly, this what that PR is about, except we generalize the case you descried at note no.1, but if it is not true or not clear, we can write it in another way so that it is a special case.

PR #427, opened 2022-10-19
[TASK] Remove running next DKG on genesis and use the proper mechanism once the session has elapsed

### LINKED ISSUE #425: [TASK] Remove running next DKG on genesis and use the proper mechanism once the session has elapsed.
**Issue summary**
We are running both the genesis and the next keygen at launch. With the new session logic, we don't need to be running the next keygen. Instead, we should execute the next keygen as we are in every session thereafter, meaning, we should wait for the session period time to elapse before generating the next key.

1xstj:
**Summary of changes**
- Removes queued keygen processing from genesis

Closes #425

### DIFF
--- a/dkg-gadget/src/worker.rs
+++ b/dkg-gadget/src/worker.rs
@@ -840,7 +840,7 @@ where
 		}
 
 		// Get the best next authorities using the keygen threshold
-		let session_id = queued.id; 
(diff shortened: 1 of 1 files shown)


drewstone (approved):
LGTM

PR #435, opened 2022-10-27
No need to persist the session information starting back up

### LINKED ISSUE #377: [BUG] Shutting down all nodes and turning back on fails to re-execute DKG protocols in the next session
**Describe the bug**
**Before next pub key signature is created**
In the middle of session 1, after the next DKG key has been generated and after shutting down all the nodes, the next pub key signature is never created (I believe it's possible that this message was started and dropped at the time of shutting nodes down). Therefore, the next key rotation never occurs.

Nonetheless, even after failing to generate the next signature, the new next authorities do not even start generating a key for round 3. The queued DKG state stays in round 2, and the DKG protocol just goes silent effectively.

**After next pub key signature is created**
After shutting down the nodes after the next pk signature is created, we see the rotation succeed, the next pk generated, but the transaction to push the next pk on-chain FAIL. This is odd as there shouldn't be an issue to inserting the key if it generated successfully.

**To Reproduce**
1. Run the DKG
2. At different points in time, shut down all collator nodes and start back up.
3. Attempt to break the protocol and see different behavior.

**Expected behaviour**
1. If we miss the signature in one session, the next key gen protocol should still execute and we should be able to update from rounds 1 -> 3 or more. Failure to handle the next authorities shouldn't stop the next next authorities from attempting to run the DKG. Since nothing happens it seems a bit odd.
2. If we miss the signature in one session, we may want to have a fallback on-chain to restart the refresh before the next session.

### LINKED ISSUE #379: [TASK] Persist the round information necessary for starting back up.
**Issue summary**
We need to persist some information about the round so that if a node shuts down and turns back on that we can initialize the worker's state with whatever round information is needed.

Related to #377

shekohex:
**Summary of changes**
- Remove Unused methods and logic for Persisting the session information necessary for starting back up.
- During the testing, shutting down the node and starting it back up again works as normal, I guess that issues are got resolved with the recent updates and improvements we did to the DKG.

- Closes #379 
- Closes #377

### DIFF
--- a/dkg-gadget/src/async_protocols/incoming.rs
+++ b/dkg-gadget/src/async_protocols/incoming.rs
@@ -92,8 +92,8 @@ impl TransformIncoming for Arc<SignedDKGMessage<Public>> {
 				}
 			},
 
-			(l, r) => {
-				log::warn!("Received message for mixed stage: Local: {:?}, payload: {:?}", l, r);
+			(_l, _r) => {
+				// log::warn!("Received message for mixed stage: Local: {:?}, payload: {:?}", l, r);
 				Ok(None)
 			},
 		}
--- a/dkg-gadget/src/persistence.rs
+++ b/dkg-gadget/src/persistence.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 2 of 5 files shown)


drewstone (approved):
LG

PR #436, opened 2022-10-27
Reorganise the logic for the offchain storage in the DKG pallet 

### LINKED ISSUE #432: [SPEC] Clear offchain storage only when the data is submitted on-chain
# Overview 
Currently we do clear the offchain storage regardless if that data is submitted on-chain or not, which is not correct, we should instead clear it if the data got submitted on-chain.

## Research
You can see example of that issue here:

1. at `submit_genesis_public_key_onchain`:
https://github.com/webb-tools/dkg-substrate/blob/083ccfa644e91247677aef30446c98587e0d5145/pallets/dkg-metadata/src/lib.rs#L1533-L1538
2. at `submit_next_public_key_onchain`:
https://github.com/webb-tools/dkg-substrate/blob/083ccfa644e91247677aef30446c98587e0d5145/pallets/dkg-metadata/src/lib.rs#L1590-L1602
3. at `submit_public_key_signature_onchain`:
https://github.com/webb-tools/dkg-substrate/blob/083ccfa644e91247677aef30446c98587e0d5145/pallets/dkg-metadata/src/lib.rs#L1633-L1646

## Examples 

We should only clear the offchain storage once we send the transaction without any errors.
A good example could be found here at `submit_misbehaviour_reports_onchain`:

https://github.com/webb-tools/dkg-substrate/blob/083ccfa644e91247677aef30446c98587e0d5145/pallets/dkg-metadata/src/lib.rs#L1685-L1696

shekohex:
**Summary of changes**
- `hex` crate now uses `alloc` feature to use it in `no_std` things, that will help us print hex values.
- Address all changes in the #432 

Closes #432

### DIFF
--- a/pallets/dkg-metadata/Cargo.toml
+++ b/pallets/dkg-metadata/Cargo.toml
@@ -19,7 +19,7 @@ codec = { package = "parity-scale-codec", version = "3", features = [
 scale-info = { version = "2.1.1", default-features = false, features = [
 	"derive",
 ] }
-hex = { version = "0.4", default-features = false }
+hex = { version = "0.4", default-features = false, features = ["alloc"] }
 libsecp256k1 = { version = "0.7.0", default-features = false }
 dkg-runtime-primitives = { path = "../../dkg-runtime-primitives", default-features = false }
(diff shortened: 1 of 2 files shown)


drewstone (approved):
LG

PR #437, opened 2022-10-31
Change Offchain worker txs to be free

### LINKED ISSUE #433: [TASK] Make all of offchain worker tx free (`submit_public_key`, `submit_next_public_key`, ..etc)
**Task summary**
Submitting Transactions for the following functions should be free and does not require any balance, could also be unsigned tx.

- [ ] [`submit_public_key`](https://github.com/webb-tools/dkg-substrate/blob/083ccfa644e91247677aef30446c98587e0d5145/pallets/dkg-metadata/src/lib.rs#L709).
- [ ] [`submit_next_public_key`](https://github.com/webb-tools/dkg-substrate/blob/083ccfa644e91247677aef30446c98587e0d5145/pallets/dkg-metadata/src/lib.rs#L764).
- [ ] [`submit_public_key_signature`](https://github.com/webb-tools/dkg-substrate/blob/083ccfa644e91247677aef30446c98587e0d5145/pallets/dkg-metadata/src/lib.rs#L815).
- [ ] [`submit_misbehaviour_reports`](https://github.com/webb-tools/dkg-substrate/blob/083ccfa644e91247677aef30446c98587e0d5145/pallets/dkg-metadata/src/lib.rs#L901).

**Specification reference**
* The way that these should work like the following example:
1. Make use of [`ValidateUnsigned`](https://docs.rs/frame-support/3.0.0/frame_support/pallet_prelude/trait.ValidateUnsigned.html) trait
```rust
#[pallet::validate_unsigned]
impl<T: Config> ValidateUnsigned for Pallet<T> {
    type Call = Call<T>;

    /// Validate unsigned call to this module.
    ///
    /// By default unsigned transactions are disallowed, but implementing the validator
    /// here we make sure that some particular calls (the ones produced by offchain worker)
    /// are being whitelisted and marked as valid.
    fn validate_unsigned(_source: TransactionSource, call: &Self::Call) -> TransactionValidity {
        // Firstly let's check that we call the right function.
        // Here we will use match stmt, to match over the call and see if it is
        // one of the functions above. if not we should return
        // `InvalidTransaction::Call.into()`

        // Next, is to validate the signature and data, depending on the function
        // we are calling, for example, for `submit_next_public_key` we should
        // validate the signature, and it is been signed by the current DKG.
        // if everything is good, we shall return `ValidTransaction`.
        //
        // See this example here:
        // https://github.com/paritytech/substrate/blob/f2bc08a3071a91b71fec63cf2b22c707411cec0e/frame/examples/offchain-worker/src/lib.rs#L298-L326
        // and here:
        // https://github.com/paritytech/substrate/blob/f2bc08a3071a91b71fec63cf2b22c707411cec0e/frame/examples/offchain-worker/src/lib.rs#L694-L719
    }
}

``` 

2. Make use of [`Signer::any_account().send_unsigned_transaction(...)` ](https://docs.rs/frame-system/3.0.0/frame_system/offchain/struct.Signer.html#method.send_unsigned_transaction).

```rust
...
    let (_, result) = Signer::<T, T::AuthorityId>::any_account()
        .send_unsigned_transaction(
            |account| {
                // Build or send the payload here.
            },
            |payload, signature| {
                // Return the call here.
                // Call::submit_next_public_key { ... }
            },
        )
        .ok_or("No local accounts accounts available.")?;
...

```

1. https://github.com/paritytech/substrate/blob/f2bc08a3071a91b71fec63cf2b22c707411cec0e/frame/examples/offchain-worker/src/lib.rs

shekohex:
**Summary of changes**
- See #433 

Closes #433

### DIFF
--- a/pallets/dkg-metadata/src/lib.rs
+++ b/pallets/dkg-metadata/src/lib.rs
@@ -112,7 +112,7 @@ use frame_support::{
 	pallet_prelude::{Get, Weight},
 	traits::{EstimateNextSessionRotation, OneSessionHandler},
 };
-use frame_system::offchain::{SendSignedTransaction, Signer};
+use frame_system::offchain::{Signer, SubmitTransaction};
 pub use pallet::*;
 use sp_runtime::{
 	generic::DigestItem,
@@ -218,6 +218,19 @@ pub mod pallet {
 		/// Percentage session should have progressed for refresh to begin
 		#[pallet::constant]
 		type RefreshDelay: Get<Permill>;
+
+		/// Number of blocks of cooldown after unsigned transaction is included.
+		///
+		/// This ensures that we only accept unsigned transactions once, every `UnsignedInterval`
+		/// blocks.
+		#[pallet::constant]
+		type UnsignedInterval: Get<Self::BlockNumber>;
+		/// A configuration for base priority of unsigned transactions.
+		///
+		/// This is exposed so that it can be tuned for particular runtime, when
+		/// multiple pallets send unsigned transactions.
+		#[pallet::constant]
+		type UnsignedPriority: Get<TransactionPriority>;
 
 		type WeightInfo: WeightInfo;
 	}
@@ -325,7 +338,14 @@ pub mod pallet {
 	#[pallet::storage]
 	#[pallet::getter(fn refresh_delay)]
 	pub type RefreshDelay<T: Config> = StorageValue<_, Permill, ValueQuery>;
-
+	/// Defines the block when next unsigned transaction will be accepted.
+	///
+	/// To prevent spam of unsigned (and unpayed!) transactions on the network,
+	/// we only allow one transaction every `T::UnsignedInterval` blocks.
+	/// This storage entry defines when new transaction is going to be accepted.
+	#[pallet::storage]
+	#[pallet::getter(fn next_unsigned_at)]
+	pub(super) type NextUnsignedAt<T: Config> = StorageValue<_, T::BlockNumber, ValueQuery>;
 	/// Check if there is a refresh in progress.
 	#[pallet::storage]
 	#[pallet::getter(fn refresh_in_progress)]
@@ -740,6 +760,9 @@ pub mod pallet {
 			}
 
 			if accepted {
+				// now increment the block number at which we expect next unsigned transaction.
+				let current_block = <frame_system::Pallet<T>>::block_number();
+				<NextUnsignedAt<T>>::put(current_block + T::UnsignedInterval::get());
 				return Ok(().into())
 			}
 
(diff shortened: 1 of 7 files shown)


standalone/runtime/src/lib.rs
```diff
@@ -594,6 +596,8 @@ impl pallet_dkg_metadata::Config for Runtime {
 	type OffChainAuthId = dkg_runtime_primitives::offchain::crypto::OffchainAuthId;
 	type NextSessionRotation = pallet_dkg_metadata::DKGPeriodicSessions<Period, Offset, Runtime>;
 	type RefreshDelay = RefreshDelay;
+	type UnsignedPriority = UnsignedPriority;
+	type UnsignedInterval = UnsignedInterval;
```
1xstj, line 600: We also need to add the `ValidateUnsigned` check to SignedExtra, like https://github.com/paritytech/polkadot/blob/e3f67317030ce6358596780370c8d221f3a65191/runtime/polkadot/src/lib.rs#L1540

pallets/dkg-metadata/src/lib.rs
```diff
@@ -1204,6 +1234,90 @@ pub mod pallet {
...
+			// 2. `submit_next_public_key`
+			// 3. `submit_public_key_signature`
+			// 4. `submit_misbehaviour_reports`.
+			// other than that we should return `InvalidTransaction::Call.into()`.
+			let is_valid_call = matches! {
+				call,
+				Call::submit_public_key { .. } |
+					Call::submit_next_public_key { .. } |
+					Call::submit_public_key_signature { .. } |
+					Call::submit_misbehaviour_reports { .. }
```
1xstj, line 1275: How about adding a check here to ensure the sender is a member of current authorities?

drewstone: We need checks like that everywhere to ensure this doesn't lead to spam attacks.

standalone/runtime/src/lib.rs
shekohex: I do believe there are no need for this check, since we do validate the unsigned transaction anyway in the pallet, and reject it before it hits the actual extrinsic, actually we do not even put it in the txpool.

pallets/dkg-metadata/src/lib.rs
shekohex: The spam attacks are prevented already by the check above that, but yes I agree, we will have to move the checks from the xt to a helper function and call it here and also in the xt itself again.

shekohex: > How about adding a check here to ensure the sender is a member of current authorities?

shekohex: I did add a check to only allow these unsigned tx to be from the offchain workers, by checking the source, if it is not local, it will be rejected before it reaches the txpool.

1xstj (approved):
💯

PR #440, opened 2022-11-02
Emergency keygen fallback

### LINKED ISSUE #438: [TASK] Manually Instruct the DKG Gadget to Run a new Keygen Protocol
**Task summary**
We need a simple way to trigger a new Keygen Protocol, manually, either by using a `sudo` call or using another way like the democracy pallet.

The main motivation behind this is that we should have an on-chain fallback mechanism in any case that the DKG stuck, or the key material got compromised somehow.

The way that should be designed is by exposing a runtime API that we do call from our gadget to determine if we shall trigger an emergency keygen protocol. This runtime call however should not depend on an On-Chain state, like a Storage, instead it should be a simple Event; the reason why is not just a boolean flag inside a storage is that first we will need a custom logic internally to determine if this the first time we check this flag or not (since we will read that storage every block, so we could end up reading and executing the same logic over and over until we flip that boolean to false), and ther other thing is that we will have to set that storage to `false` after running the protocol to the end, which is also more complex that it seems, hence an Event that is `fire-and-forget` is better, since we will only read it once, and if it failed the first time we can trigger another one, easily!

**Specification reference**
- Expose a runtime API that returns a boolean to determine if we should execute an emergency keygen.
- A simple extrinsic call that will emit an Event which will be read by that runtime API and filter by that event, at a sepcific block.
- At every finalized block notification, inside the gadget, call that runtime API function to determine if we shall execute that emergency keygen protocol.

1. Reading Events could be from the `System::events()` in the system pallet.

**SPEC Updates**

Looks like reading the whole system events is costly, instead we will use the other idea of on-chain storage flag to store a boolean value to indecate if we shall execute an emergency keygen protocol or not. The trick here to avoid the complex implementation of the on-chain storage, is we set it to `true` but next block we set it back to `false`.

shekohex:
**Summary of changes**
- SPEC #438 

Closes #438

### DIFF
--- a/dkg-gadget/src/worker.rs
+++ b/dkg-gadget/src/worker.rs
@@ -850,6 +850,12 @@ where
 		debug!(target: "dkg_gadget::worker", "🕸️  Processing block notification for block {}", header.number());
 		*self.latest_header.write() = Some(header.clone());
 		debug!(target: "dkg_gadget::worker", "🕸️  Latest header is now: {:?}", header.number());
+		// Check if we should execute emergency Keygen Protocol.
+		if self.should_execute_emergency_keygen(header) {
+			debug!(target: "dkg_gadget::worker", "🕸️  Should execute emergency keygen");
+			self.handle_emergency_keygen(header);
+			return
+		}
 		// Clear offchain storage
 		listen_and_clear_offchain_storage(self, header);
 		// Attempt to enact new DKG authorities if sessions have changed
@@ -1090,6 +1096,13 @@ where
 				reason: "Message signature is not from a registered authority or next authority"
 					.into(),
 			})
+		}
+	}
+
+	pub fn handle_emergency_keygen(&self, header: &B::Header) {
+		// Start the queued DKG setup for the new queued authorities
+		if let Some((_active, queued)) = self.validator_set(header) {
+			self.handle_queued_dkg_setup(header, queued);
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


dkg-gadget/src/worker.rs
```diff
@@ -850,6 +850,12 @@ where
 		debug!(target: "dkg_gadget::worker", "🕸️  Processing block notification for block {}", header.number());
 		*self.latest_header.write() = Some(header.clone());
 		debug!(target: "dkg_gadget::worker", "🕸️  Latest header is now: {:?}", header.number());
+		// Check if we should execute emergency Keygen Protocol.
+		if self.should_execute_emergency_keygen(header) {
+			debug!(target: "dkg_gadget::worker", "🕸️  Should execute emergency keygen");
+			self.handle_emergency_keygen(header);
```
1xstj, line 856: QQ : Will it be ok if a queued or current DKG is already in process and we trigger emergency keygen?

drewstone (changes requested):
Looks good, we should have tests of the flag flipping as well @shekohex

dkg-gadget/src/worker.rs
shekohex: There is handling already for that Edge case, basicly it will just ignore that call and return.
https://github.com/webb-tools/dkg-substrate/blob/4eaee85a2253e95a5c54dd29846bf2c298445d75/dkg-gadget/src/worker.rs#L794-L804

PR #442, opened 2022-11-02
[TASK] Increase Off-chain storage readability

### LINKED ISSUE #439: [TASK] Increase Off-chain storage readability
**Task summary**
A previous bug got reported and resolved by the #432 SPEC and the PR linked to it, which is better now, but we need more readability at the gadget layer which is task about.

**Specification reference**
- At every finalized block notification, we run some checks like the following:
1. Check if the offchain storage of value X is empty, and
2. Check if the corresponding value of X on is **not** on chain, also 
3. Check if `N` blocks has passed since we wrote that value on the offchain storage.

If the above checks passed, we should try to write the value again to the offchain storage so that the offchain worker try again to submit it on chain.

**Example**: Imagine we did write the `NextDkgPublicKey` to the offchain storage to be read by the offchain worker and to be submitted on chain, currently if that transaction failed, we will keep this value and try later (after #432 got implemented), but it could be in rare case that the value gets deleted from the offchain storage before we submit it on chain, in that case, the DKG gadget should try to write the same value again by running the above checks in the SPEC.

1xstj:
**Summary of changes**

Closes #439

### DIFF
--- a/dkg-gadget/src/storage/clear.rs
+++ b/dkg-gadget/src/storage/clear.rs
@@ -81,7 +81,7 @@ pub(crate) fn listen_and_clear_offchain_storage<B, BE, C, GE>(
 	}
 }
 
-/// cleans offchain storage at interval
+/// Checks if the Off-chain storage stored at `storage_key` is empty or not.
 pub(crate) fn is_offchain_storage_empty<B, BE, C, GE>(
 	dkg_worker: &DKGWorker<B, BE, C, GE>,
 	storage_key: &[u8],
@@ -98,5 +98,5 @@ where
 		return offchain.unwrap().get(STORAGE_PREFIX, storage_key).is_some()
 	}
 
-	return true
+	true
 }
--- a/dkg-gadget/src/storage/clear.rs
+++ b/dkg-gadget/src/storage/clear.rs
@@ -80,3 +80,23 @@ pub(crate) fn listen_and_clear_offchain_storage<B, BE, C, GE>(
 		}
 	}
 }
+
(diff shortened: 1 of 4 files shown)


dkg-gadget/src/storage/clear.rs
```diff
@@ -80,3 +80,23 @@ pub(crate) fn listen_and_clear_offchain_storage<B, BE, C, GE>(
 		}
 	}
 }
+
+/// cleans offchain storage at interval
```
shekohex, line 84:
```suggestion
/// Checks if the Off-chain storage stored at `storage_key` is empty or not.
```

dkg-gadget/src/worker.rs
```diff
@@ -1099,6 +1107,34 @@ where
...
+	/// being written onchain. This function will write the key again to the offchain storage so
+	/// that its pickedup by the offchain workers and tried again.
+	pub fn ensure_generated_keys_are_posted_onchain(
+		&self,
+		latest_header: &B::Header,
+		session_id: SessionId,
+	) {
+		// nextPublicKey
+		if self.get_next_dkg_pub_key(latest_header).is_none() &&
+			is_offchain_storage_empty(self, AGGREGATED_PUBLIC_KEYS)
```
shekohex, line 1123: I would also add another 3rd check which is to check if `now - self.next_rounds.start_at > N blocks` where `N=5` for example.

shekohex (changes requested):
Looks good, just needs one more check there.

shekohex (approved):

PR #445, opened 2022-11-09
[PATCH] fix for DKG Rotation and Nonce

shekohex:
**Summary of changes**
- Instead of doing the actual RefreshProposal inside the `submit_next_public_key` xt, and also inside the `on_initialize` hook, which could lead to double execution of the same proposal, instead we only now doing it in the `on_initialize` hook.
- However, we do reset the `RefreshInProgress` flag, whenever we get a new Public Key, so that the next call to `on_initialize` can call/execute the `do_refresh` call depending on the session progress.

Closes


drewstone (approved):
LGTM

PR #450, opened 2022-11-24
[feat] dkg prometheus metrics

1xstj:
**Summary of changes**
- Expose dkg metrics to prometheus

Related to https://github.com/webb-tools/tangle/issues/85

### DIFF
--- a/dkg-gadget/src/metrics.rs
+++ b/dkg-gadget/src/metrics.rs
@@ -18,9 +18,20 @@ use prometheus::{register, Counter, Gauge, PrometheusError, Registry, U64};
 /// DKG metrics exposed through Prometheus
 #[derive(Clone)]
 pub(crate) struct Metrics {
-	pub propagated_messages: Counter<U64>,
+	/// Count to total propogated messages
+	pub dkg_propagated_messages: Counter<U64>,
 	/// Current active validator set id
 	pub dkg_validator_set_id: Gauge<U64>,
+	/// Total messages received
+	pub dkg_inbound_messages: Counter<U64>,
+	/// Total error messages received
+	pub dkg_error_counter: Counter<U64>,
+	/// Current session progress witnessed by dkg worker
+	pub dkg_session_progress: Gauge<U64>,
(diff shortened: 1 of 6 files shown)


dkg-gadget/src/metrics.rs
```diff
@@ -18,9 +18,20 @@ use prometheus::{register, Counter, Gauge, PrometheusError, Registry, U64};
...
+	/// Total messages received
+	pub dkg_inbound_messages: Counter<U64>,
+	/// Total error messages received
+	pub dkg_error_counter: Counter<U64>,
+	/// Current session progress witnessed by dkg worker
+	pub dkg_session_progress: Gauge<U64>,
+	/// Keygen retry counter for dkg worker
+	pub dkg_keygen_retry_counter: Counter<U64>,
+	/// The latest block height witnessed by the dkg worker
+	pub dkg_latest_block_height: Gauge<U64>,
```
dutterbutter, line 34: Should we add a metrics for key rotation or is that captured in session progress / keygen retry?

dkg-gadget/src/metrics.rs
```diff
@@ -30,10 +41,36 @@ impl Metrics {
...
 				registry,
 			)?,
+			dkg_inbound_messages: register(
+				Counter::new("dkg_inbound_messages", "Number of DKG messages received.")?,
+				registry,
+			)?,
+			dkg_error_counter: register(
+				Counter::new("dkg_error_counter", "Number of DKG errors generated")?,
+				registry,
+			)?,
```
dutterbutter, line 55: Will this provide insights into what dkg error? For instance, will it assist in detecting when a dkg keygen fails or just report an error of 1. Idk if that is exactly clear lol

1xstj: It will report how many times the `handle_dkg_error` function has been triggered, not all are critical. I think we can only track metrics here, we should rely on logs for the exact error.

dutterbutter (approved):

PR #470, opened 2022-12-28
DKG database backend

### LINKED ISSUE #468: [SPEC] Re:architect How we store the DKG Worker State
## Overview 

The Current DKG State is stored in two places, the on-chain storage and the off-chain storage. but in the off-chain storage using file based storage, where we for example store the local key share encrypted on disk, there are a lot of branches and flows that makes it more complicated to work with, and also susceptible to data corruption.  

## Research

There are one solution to this, by using some sort of a database, and it should be easy to use, for that there are two options:
1. [sled](https://sled.rs) - it is good, we have been using it for the relayer since forever, and it is easy to use and robust.
2. [Substrate off-chain storage](https://docs.substrate.io/reference/how-to-guides/offchain-workers/offchain-local-storage/) Built-in, not that hard to use, but it will be sufficient to get the work done.

### Using sled:

* Pros
	- Easy to use
	- Single file database, easy backup.
	- We can implement an Encryption Layer on top.
* Cons
	- we will keep using and requiring the nodes to run with a custom `base-path`, see #288 
	- They say as the data gets so big, over the time, sled used more memory, see https://github.com/spacejam/sled/issues/1303

### Using off-chain storage

* Pros
	- Built-in, no need to get any external code.
	- Same infra is used on the on-chain storage, so it is well-tested and trusted.
	- We will get rid of depending on the `base-path` option.
	- we can implement the encryption layer on top of that easily. 
* Cons
	- None?

## Architecture

There are a few areas that we shall revisit for this new design for the state of the DKG:

- Storing the DKG `LocalKey` share, currently we store it in a files in [`persistence.rs`](https://github.dev/webb-tools/dkg-substrate/blob/82ebc7dccd80b7ba6cde96471e23d9fc4a913105/dkg-gadget/src/persistence.rs)  

A simple trait (for now) that can be extended as we go, could look something like the following:

```rust
/// A Database backend, specificly for the DKG to store and load important state
///
/// The backend of this database could be using a persistence store, or in-memory
/// ephermal store, depending on the use case, for example during the tests we can switch
/// to in-memory store, and in production we could use sled database or Offchain storage.
trait DKGDbBackend {
    type Error;
    /// Returns the DKG [`LocalKey<Secp256k1>`] at specific session, if any.
    fn get_local_key(
        &self,
        session_id: SessionId,
    ) -> Result<Option<LocalKey<Secp256k1>>, Self::Error>;
    /// Stores the [`LocalKey<Secp256k1>`] at this session.
    fn store_local_key(
        &self,
        session_id: SessionId,
        local_key: LocalKey<Secp256k1>,
    ) -> Result<(), Self::Error>;
}
```

shekohex:
**Summary of changes**
- See #468 
- For the seed used in generating the signing sets, we now using the on-chain public key as the seed, instead of the local one.
- we do now shutdown and clear all signing rounds, once we rotate sessions locally.
- there is no need now for providing the `base_path`.
- `fetch_local_keys` now takes an optional session id, to fetch the local keys at specific session, this useful in case that the node did shutdown and there is no `rounds` nor `next_rounds`, fetching the current active session from on-chain would help in such cases.
- we now prevent the bug where we do generate a new next DKG key, while there is already one that is submitted on-chain.

Closes #468


drewstone:
E2E tests don't seem to run. Can you try and fix this @shekohex ?

dkg-gadget/src/db/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+mod mem;
+mod offchain_storage;
+
+pub use mem::DKGInMemoryDb;
+pub use offchain_storage::DKGOffchainStorageDb;
+
+/// A Database backend, specificly for the DKG to store and load important state
+///
+/// The backend of this database could be using a persistence store, or in-memory
+/// ephermal store, depending on the use case, for example during the tests we can switch
```
drewstone, line 14:
```suggestion
/// ephemeral store, depending on the use case. For example, during the tests we can switch
```

dkg-gadget/src/db/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+
+mod mem;
+mod offchain_storage;
+
+pub use mem::DKGInMemoryDb;
+pub use offchain_storage::DKGOffchainStorageDb;
+
+/// A Database backend, specificly for the DKG to store and load important state
+///
+/// The backend of this database could be using a persistence store, or in-memory
```
drewstone, line 13:
```suggestion
/// The backend of this database could be using a persistence store or in-memory
```

dkg-gadget/src/db/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+/// ephermal store, depending on the use case, for example during the tests we can switch
+/// to in-memory store, and in production we could use sled database or Offchain storage.
```
drewstone, line 15:
```suggestion
/// to an in-memory store, and in production we could use sled database or Offchain storage.
```

dkg-gadget/src/db/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+#[auto_impl::auto_impl(Arc, &, &mut)]
+pub trait DKGDbBackend: Send + Sync + 'static {
+	/// Returns the DKG [`LocalKey<Secp256k1>`] at specific session, if any.
```
drewstone, line 18:
```suggestion
	/// Returns the DKG [`LocalKey<Secp256k1>`] at a specific session or an error if the session doesn't exist.
```

dkg-gadget/src/db/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+	fn get_local_key(&self, session_id: SessionId)
+		-> Result<Option<LocalKey<Secp256k1>>, DKGError>;
+	/// Stores the [`LocalKey<Secp256k1>`] at this session.
```
drewstone, line 21:
```suggestion
	/// Stores the [`LocalKey<Secp256k1>`] at a specified session.
```

dkg-gadget/src/db/offchain_storage.rs
```diff
@@ -0,0 +1,223 @@
+//! DKG Database backend, implemented using Offchain Storage.
```
drewstone, line 1:
```suggestion
//! DKG Database backend, implemented using Substrate's Offchain Storage.
```

dkg-gadget/src/db/offchain_storage.rs
```diff
@@ -0,0 +1,223 @@
...
+//! Unlike the in-memory database, this database is persistent, and can be used to store
```
drewstone, line 2:
```suggestion
//! Unlike the in-memory database, this database is persistent and can be used to store
```

dkg-gadget/src/db/offchain_storage.rs
```diff
@@ -0,0 +1,223 @@
...
+use sc_keystore::LocalKeystore;
+use sp_core::{offchain::OffchainStorage, Pair};
+use sp_runtime::traits::Block;
+
+use crate::DKGKeystore;
+
+/// DKG Offchain Storage prefix.
+const STORAGE_PREFIX: &[u8] = b"dkg";
+
+/// A Database backend, specificly for the DKG to store and load important state
```
drewstone, line 25:
```suggestion
/// A Database backend, specifically for the DKG to store and load important state
```

dkg-gadget/src/db/offchain_storage.rs
```diff
@@ -0,0 +1,223 @@
...
+/// implemented using Offchain Storage.
+///
+/// This backend also uses the DKG Keystore to stores the data in an encrypted form.
```
drewstone, line 28:
```suggestion
/// This backend also uses the DKG Keystore to store the data in an encrypted form.
```

dkg-gadget/src/lib.rs
```diff
@@ -176,15 +171,23 @@ where
 	let keygen_handle = tokio::spawn(keygen_gossip_handler.run());
 	let signing_handle = tokio::spawn(signing_gossip_handler.run());
 
+	// In memory backend, not used for now
+	// let db_backend = Arc::new(db::DKGInMemoryDb::new());
```
drewstone, line 175: Do you think we should extract this into a CLI flag? What is the benefit of using in memory.. is it just for testing? Seems like we should be able to test with offchain storage directly and simplify our approaches. No need to carry around extra baggage imo.

dkg-gadget/src/worker.rs
```diff
@@ -516,57 +511,22 @@ where
...
-			} else {
-				debug!("Successfully copied queued key to current key");
-				return true
-			}
-		}
-
-		false
-	}
-
 	/// Fetch the local stored keys if they exist.
```
drewstone, line 514:
```suggestion
	/// Fetch the stored local keys if they exist.
```

drewstone (changes requested):
LG, please fix up the CI, get the E2E tests and unit tests fixed with the updated values.

PR #471, opened 2023-01-09
SBP Milestone 1 review

Wiezzel:
Summary:
1. Usage of `without_storage_info` [is deprecated](https://github.com/paritytech/substrate/issues/8629).
    Removing this will require replacing `Vec` with `BoundedVec` in storage items.
2. Some docstrings seem copied from another codebase and don't match the code.
3. There are some operations that might potentially panic.
4. Extrinsics are now transactional by default, no need for `#[transactional]`
    See https://github.com/paritytech/substrate/pull/11431
5. There are some unsafe arithmetic operations. Use of `saturating_` or `checked_` operations is recommended instead.


drewstone:
Thanks for this @Wiezzel ! I will address these this week.

drewstone:
@1xstj I've closed this since the completion email but do double check we've addressed all these items.

PR #474, opened 2023-01-12
[fix] add logs for key fetching

1xstj:
**Summary of changes**
- Add logs to help identify the current session_id and the current key being used

### DIFF
--- a/.github/workflows/image-publish.yml
+++ b/.github/workflows/image-publish.yml
@@ -24,10 +24,6 @@ jobs:
       - uses: actions/checkout@v2
       - name: Build Image
         run: ./scripts/build-standalone-docker.sh
-      - name: Install docker-slim
-        run: curl -sL https://raw.githubusercontent.com/slimtoolkit/slim/master/scripts/install-slim.sh | sudo -E bash -
-      - name: Slim down the Image
-        run: slim build --target webb-tools/dkg-standalone-node --http-probe-off --entrypoint dkg-standalone-node
       - name: Log into registry
         run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
       - name: Push image
@@ -48,5 +44,5 @@ jobs:
 
           echo IMAGE_ID=$IMAGE_ID
           echo VERSION=$VERSION
-          docker tag webb-tools/dkg-standalone-node.slim $IMAGE_ID:$VERSION
+          docker tag webb-tools/dkg-standalone-node $IMAGE_ID:$VERSION
           docker push $IMAGE_ID:$VERSION


dkg-gadget/src/async_protocols/blockchain_interface.rs
```diff
@@ -202,6 +202,7 @@ where
 		key: LocalKey<Secp256k1>,
 		session_id: SessionId,
 	) -> Result<(), DKGError> {
+		log::info!(target: "dkg", "Storing public key {:?} for session {:?}", key, session_id);
```
shekohex, line 205: This a bit redundant, I would say a better way is to keep all these logs inside the database backend.

dkg-gadget/src/db/offchain_storage.rs
```diff
@@ -72,12 +72,14 @@ where
...
 		session_id: SessionId,
 	) -> Result<Option<LocalKey<Secp256k1>>, DKGError> {
+		log::info!(target: "dkg", "Offchain Storage : Fetching local keys for session {:?}", session_id);
 		let db_key = keys::LocalKey::new(session_id);
 		let maybe_decrypted_bytes = self.load_and_decrypt(codec::Encode::encode(&db_key))?;
 		match maybe_decrypted_bytes {
 			Some(decrypted_bytes) => {
 				let local_key = serde_json::from_slice(&decrypted_bytes.0)
 					.map_err(|e| DKGError::CriticalError { reason: e.to_string() })?;
+				log::info!(target: "dkg", "Offchain Storage : Fetched local keys for session {:?}, Key : {:?}", session_id, local_key);
```
shekohex, line 82: Printing the full `local_key` is not a good option here, it contains a lot of other information that is not important, we only interested in the `public_key` field, and in hex format which is great for us to compare.

dkg-gadget/src/db/offchain_storage.rs
```diff
@@ -206,6 +209,9 @@ where
 		let mut offchain_storage = self.backend.offchain_storage().ok_or_else(|| {
 			DKGError::CriticalError { reason: String::from("No Offchain Storage available!!") }
 		})?;
+		if let Some(_) = offchain_storage.get(STORAGE_PREFIX, &key) {
+			log::warn!("Offchain Storage : Overwriting already existing key for session {:?}", key);
+		}
```
shekohex, line 214:
```suggestion
		if let Some(_) = offchain_storage.get(STORAGE_PREFIX, &key) {
			log::warn!("Offchain Storage : Overwriting already existing database entry at key 0x{}", hex::encode(key));
		}
```

dkg-gadget/src/worker.rs
```diff
@@ -522,10 +522,13 @@ where
 	) -> (Option<LocalKey<Secp256k1>>, Option<LocalKey<Secp256k1>>) {
 		let current_session_id =
 			self.rounds.read().as_ref().map(|r| r.session_id).or(optional_session_id);
+		debug!(target: "dkg_gadget::worker", "🕸️  Fetching local keys for session id : {:?}", current_session_id);
 		let next_session_id = current_session_id.map(|s| s + 1);
 		let active_local_key =
 			current_session_id.and_then(|s| self.db.get_local_key(s).ok().flatten());
 		let next_local_key = next_session_id.and_then(|s| self.db.get_local_key(s).ok().flatten());
+		debug!(target: "dkg_gadget::worker", "🕸️  Fetched active-local-key : {:?}", active_local_key);
+		debug!(target: "dkg_gadget::worker", "🕸️  Fetched next-local-key : {:?}", next_local_key);
```
shekohex, line 531: It is better to only print the `hex` value of the `public_key`.

shekohex (changes requested):

shekohex:
Can we fix the lints first? so that CI passes

shekohex (approved):
LGTM!

PR #475, opened 2023-01-12
[feat] bump substrate deps to v0.9.36

1xstj:
**Summary of changes**
- Bump all deps to v0.9.36


drewstone:
I think runtime weights may need to be recalculated @1xstj as per the errors in CI.

PR #479, opened 2023-01-17
Fixes to the Signing Rounds

### LINKED ISSUE #478: [BUG] Unsigned Proposals are not removed from queue after signing
**Describe the bug**
The signed proposals are not correctly removed from the unsignedProposals queue

**To Reproduce**
1. `cargo b -rp dkg-standalone-node`
2. Run dkg standalone node with 3 validators
3. As the dkg progresses, we can see that the unsignedProposal queue grows

**Expected behaviour**
The unsigned proposal queue is emptied on proposal expiry or signing

shekohex:
**Summary of changes**
- Validate that the status of the `AsyncProtocolRemote` is being set in the correct order.
- check the state if it is active so that we can send a shutdown signal.
- **Allow Overwriting signing rounds.** (Hopefully fixes the `InvalidKey` in the MP-ECDSA)

Closes #478

### DIFF
--- a/dkg-gadget/src/async_protocols/incoming.rs
+++ b/dkg-gadget/src/async_protocols/incoming.rs
@@ -87,7 +87,7 @@ impl TransformIncoming for Arc<SignedDKGMessage<Public>> {
 						Ok(None)
 					}
 				} else {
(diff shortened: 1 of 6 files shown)


drewstone (approved):
LGTM

Issue #494, opened 2023-02-16
[TASK] Verify and Filter incoming messages at the `StateMachineWrapper` level


shekohex:
**Task summary**

We do see a lot of `MsgOverwrite` errors in our Async Protocol, so we suspect it is an issue that one peer sends us the same message twice. once the inner state machine receives the same message twice, it bails and returns the error [here](https://github.com/ZenGo-X/round-based-protocol/blob/959126f9f6edce16d4ee95954091b93e33a83140/src/containers/broadcast.rs#L129-L131).

One possible solution is the following:

1. We check and filter out the messages if we have seen it before.
2. We do match over the error inside [`AsyncPorotocl`](https://github.com/ZenGo-X/round-based-protocol/blob/959126f9f6edce16d4ee95954091b93e33a83140/src/async_runtime/mod.rs#L52) and inside the method [`new_inner`](https://github.com/webb-tools/dkg-substrate/blob/22084634b91a6e92c3b55e432e1bdea8fedf8fc7/dkg-gadget/src/async_protocols/mod.rs#L271-L281) and if we found `MsgOverwrite` error, we skip and do not finish the state machine.

**Other information and links**

PR #495, opened 2023-02-16
Handle Critical and Non-Critical Errors in the Async Protocol 

shekohex:
**Summary of changes**
- previously, we handled all the async protocol errors are critical errors, but now we differentiate between critical and non critical errors.  

Related to #494

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5216,7 +5216,7 @@ dependencies = [
 [[package]]
 name = "multi-party-ecdsa"
 version = "0.8.2"
-source = "git+https://github.com/webb-tools/multi-party-ecdsa.git#574b863105e9c0bd796c866057764e0e55a49d54"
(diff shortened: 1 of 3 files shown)


drewstone:
To that extent, we could even filter for the specific errors we believe are not critical @shekohex such as the `MsgOverwrite` and `ReceivedOutOfOrderMessage` until we get something pushed upstream. If that fixes our issues then even better, we'd know that they are not critical and can push that change upstream in a PR and update accordingly.

shekohex:
> To that extent, we could even filter for the specific errors we believe are not critical @shekohex such as the `MsgOverwrite` and `ReceivedOutOfOrderMessage` until we get something pushed upstream. If that fixes our issues then even better, we'd know that they are not critical and can push that change upstream in a PR and update accordingly.

Yes, the PR for that already up here: https://github.com/webb-tools/multi-party-ecdsa/pull/11

PR #501, opened 2023-02-17
Read `party_i` from on-chain storage

shekohex:
**Summary of changes**
- Reads `party_i` from the on chain storage instead of recalculating it.
- Prints the party index for easier debugging.

Closes


drewstone (approved):
LG

PR #502, opened 2023-02-17
Remove the wrong assumption about `party_i` in Offline Stages

shekohex:
**Summary of changes**
- Removed the `assert_eq!` check.

> OfflineStages takes a different party_i than the one used in the keygen, there is one additional index that is used in generating OfflineStages, for example, we may choose signers like `{1,3,4,5}` and if my `party_index = 1`, then there’s another index (`offline_i`) I use which is `0` (in that case).


drewstone:
Can you change the naming of this parameter here so we don't confuse them? Please refer to what it is in MP-ECDSA lib.

PR #505, opened 2023-02-21
fix signing set metric & set MAX_SIGNING_SETS to 8

1xstj:
**Summary of changes**
- Set metric after value is updated

Closes


shekohex:
Could we also adjust the maximum signing sets count from `32` to just `8`.

shekohex (approved):

PR #510, opened 2023-02-28
Improve the DKG Networking by prefer the p2p communication when possible

shekohex:
**Summary of changes**
- [x] Add The Handshake message
- [x] Verify the Handshake messages
- [x] Store and create the mapping between `PeerId` `<=>` `AuthorityId`
- [x] Introduce `KeygenPartyId`, `OfflinePartyId` and mapping between both of them.
- [x] Test utils functions for converting between `KeygenPartyId` to `OfflinePartyId` and back, with respect to the `AuthorityId`.
- [x] Introduce the `recipient_id` inside the DKG Messages.
- [x] Implement the Peer to Peer messaging.

**Noticiable Changes**

1. No more direct use of `u16` as party ids, we either use `KeygenPartyId` or `OfflinePartyId` depending on the context, to catch any bugs at compile-time. There is a direct conversation between both types and tests to ensure that these are correct.
2. When a peer connects to us, we send them our `Handshake`, and they will also send their `Handshake` message to us, this message (as of now) contains few information which includes their `AuthorityId`, which is signed and verified. Any invalid handshake message will result in a report of that peer and a disconnection of this imposter. Note that `Handshake` messages should never be broadcasted to other people.
3. MP-ECDSA Messages may contain a specific receiver value, where that value is in range of `[1; n]` this value is then mapped to `AuthorityId`, from there, we can obtain the `PeerId` of that authority, and we send them the message directly, when possible, otherwise, we fallback to gossiping that message.

**Future Ideas:**
> 1. Gossip the known peers to the whole network, build a local graph of each peer in the network, find the shortest path to deliver the message though the peers.

> 2. while building the signing sets, prefer to pick up the known peers over the unknown peers.
  

Closes

### DIFF
--- a/dkg-gadget/src/async_protocols/mod.rs
+++ b/dkg-gadget/src/async_protocols/mod.rs
@@ -194,7 +194,7 @@ impl fmt::Display for KeygenPartyId {
 
 impl KeygenPartyId {
 	/// Try to convert a KeygenPartyId to an OfflinePartyId.
-	pub fn try_to_offline_party_id(&self, s_l: &[u16]) -> Result<OfflinePartyId, DKGError> {
+	pub fn try_to_offline_party_id(&self, s_l: &[Self]) -> Result<OfflinePartyId, DKGError> {
 		OfflinePartyId::try_from_keygen_party_id(*self, s_l)
 	}
 
@@ -227,19 +227,21 @@ impl OfflinePartyId {
 	/// This is safe because the KeygenPartyId is guaranteed to be in the range `[1, n]`, and the
 	/// OfflinePartyId is guaranteed to be in the range `[1, t+1]`. if the KeygenPartyId is not in
 	/// the list of signing parties, then we return an error.
-	pub fn try_from_keygen_party_id(i: KeygenPartyId, s_l: &[u16]) -> Result<Self, DKGError> {
-		let party_i = *i.as_ref();
+	pub fn try_from_keygen_party_id(
+		i: KeygenPartyId,
+		s_l: &[KeygenPartyId],
+	) -> Result<Self, DKGError> {
 		// find the index of the party in the list of signing parties
-		let index = s_l.iter().position(|&x| x == party_i).ok_or(DKGError::InvalidKeygenPartyId)?;
+		let index = s_l.iter().position(|&x| x == i).ok_or(DKGError::InvalidKeygenPartyId)?;
 		let offline_id = index as u16 + 1;
 		Ok(Self(offline_id))
 	}
 
 	/// Try to Converts the `OfflinePartyId` to `KeygenPartyId`.
-	pub fn try_to_keygen_party_id(&self, s_l: &[u16]) -> Result<KeygenPartyId, DKGError> {
+	pub fn try_to_keygen_party_id(&self, s_l: &[KeygenPartyId]) -> Result<KeygenPartyId, DKGError> {
 		let idx = self.to_index();
 		let party_i = s_l[idx];
-		KeygenPartyId::try_from(party_i)
+		Ok(party_i)
 	}
 
 	/// Converts the OfflinePartyId to an index.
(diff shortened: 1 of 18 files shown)


dkg-gadget/src/async_protocols/mod.rs
```diff
@@ -143,6 +143,113 @@ impl CurrentRoundBlame {
...
+		i: KeygenPartyId,
+		s_l: &[KeygenPartyId],
+	) -> Result<Self, DKGError> {
+		// find the index of the party in the list of signing parties
+		let index = s_l.iter().position(|&x| x == i).ok_or(DKGError::InvalidKeygenPartyId)?;
+		let offline_id = index as u16 + 1;
+		Ok(Self(offline_id))
+	}
+
+	/// Try to Converts the `OfflinePartyId` to `KeygenPartyId`.
```
tbraun96, line 240:
```suggestion
	/// Tries to convert the `OfflinePartyId` to a `KeygenPartyId`.
```

dkg-gadget/src/async_protocols/mod.rs
```diff
@@ -143,6 +143,113 @@ impl CurrentRoundBlame {
...
+	pub fn try_to_keygen_party_id(&self, s_l: &[KeygenPartyId]) -> Result<KeygenPartyId, DKGError> {
+		let idx = self.to_index();
+		let party_i = s_l[idx];
```
tbraun96, line 243: Since this function is fallible, we should take advantage of that and not index directly into `s_l`. Try `s_l.get(idx).ok_or(..)`

tbraun96 (commented):
The code looks solid to me. Just a few nits

dkg-gadget/src/gossip_engine/network.rs
```diff
@@ -502,6 +520,12 @@ impl<B: Block + 'static> GossipHandler<B> {
 				if protocol == self.protocol_name =>
 			{
 				debug!(target: "dkg_gadget::gossip_engine::network", "Peer {} connected to gossip protocol", remote);
+				// Send our Handshake message to that peer.
+				if let Err(err) = self.send_handshake_message(remote).await {
```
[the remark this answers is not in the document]
shekohex: That's true, but it is already guarded by `if protocol == self.protocol_name`, means that non-validator nodes will not start the DKG Gadget in the first place, hence they will not start that gossip engine code.

dkg-gadget/src/gossip_engine/network.rs
```diff
@@ -520,9 +548,23 @@ impl<B: Block + 'static> GossipHandler<B> {
...
-				debug_assert!(_peer.is_some());
+				let peer = lock.remove(&remote);
+				debug_assert!(peer.is_some());
+				// remove that peer's authority id from the authority id to peer id map (if any).
+				match peer {
+					Some(peer) =>
+						if let Some(authority_id) = peer.authority_id {
+							let expected_remote =
+								self.authority_id_to_peer_id.write().remove(&authority_id);
+							// This should always be valid, if it isn't, it means that we have a bug
```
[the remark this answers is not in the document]
shekohex: same as above, these notifications/messages are only per protocol, not the global connections.

dkg-gadget/src/async_protocols/mod.rs
shekohex: Good catch here, will add that check.

dkg-gadget/src/gossip_messages/dkg_message.rs
```diff
@@ -54,6 +54,10 @@ pub(crate) fn sign_and_send_messages<GE>(
 
 				crate::utils::inspect_outbound(ty, encoded_signed_dkg_message.len());
 
+				// This method name here is a little bit misleading, as it could actually be a peer
+				// to peer message not a gossiping, depends on whether the message contains a
+				// recipient or not. So the it is up to the underlying gossip engine to decide
```
drewstone, line 59:
```suggestion
				// recipient or not. So it is up to the underlying gossip engine to decide
```

drewstone (approved):

PR #520, opened 2023-03-09
Add bridges list in genesis config (fixes #515)

### LINKED ISSUE #515: [TASK] Add `BridgeMetadata` list in the Genesis config
**Task summary**

We need to include the initial Bridges in the genesis config files, for that we need to add support for adding bridges from the `GenesisConfig` for the bridge-registry pallet

**Specification**
Right now, the `GenesisConfig` does not support adding `BridgeMetadata` from the config itself

https://github.com/webb-tools/dkg-substrate/blob/052353713e4e855a512897818c66fac2615c5dbe/pallets/bridge-registry/src/lib.rs#L107-L124

To support this, we need to change the struct to something like the following:

```rust
	pub struct GenesisConfig<T: Config<I>, I: 'static = ()> {
		pub bridges: Vec<BridgeMetadata<T::MaxResources, T::MaxAdditionalFields>>,
		pub _marker: PhantomData<I>,
	}
```

That way we can register bridges from the chain_spec

https://docs.substrate.io/reference/how-to-guides/basics/configure-genesis-state/

Nutomic:
**Summary of changes**
- Adds bridges parameter to bridge-registry GenesisConfig which allows adding bridges at build time
- Add a testing bridge in `testnet_genesis()` chainspec
- Fix for crash in `submitProposals.ts` (@shekohex)
- Use existing `ResourceId` struct instead of typedef
- Use cargo run in `scripts/run-standalone.sh` instead of executing binary, so that separate build step is not necessary
- Extract some constants from chainspec code

Closes #515


standalone/node/src/chain_spec.rs
```diff
@@ -380,7 +370,17 @@ fn testnet_genesis(
...
+			bridges: bounded_vec![BridgeMetadata {
+				resource_ids: bounded_vec![RESOURCE_ID_HERMES_ATHENA, RESOURCE_ID_ATHENA_HERMES],
+				info: BridgeInfo {
+					additional: Default::default(),
+					display: SerdeData::from_str("hermes-athena")
+						.unwrap()
+				}
+			}],
+			..Default::default()
+		},
```
Nutomic, line 383: Here are the bridge parameters being passed in. This works and shows up as expected in polkadot web ui. However the test case in https://github.com/webb-tools/relayer/pull/355 is failing now with `Metadata(IncompatibleStorageMetadata("BridgeRegistry", "Bridges"))`, I think its necessary to update the dependency in webb-rs crate to fix it.

shekohex: Yup, this needs an Update to webb-rs, can you generate the new metadata locally and use the local version of webb-rs to test that?

shekohex (approved):

shekohex:
Needs to fix the CI, and this good to be merged!

standalone/node/src/chain_spec.rs
Nutomic: I did that according to instructions, files `src/substrate/dkg_runtime.rs` and `src/substrate/protocol_substrate_runtime.rs` got updated. I changed the relayer webb dependency to use the local repo path, however its still giving the same error. Not sure what Im missing.

PR #561, opened 2023-04-12
Move @polkadot/dev to dev dependencies to support Node 18

Nutomic:
**Summary of changes**
- Moves `@polkadot/dev` package to dev dependencies, so that relayer tests can be built with Node 18. Im not sure how to test if this is actually enough, or if more changes are needed.

I also tried to upgrade this package and make it build with Node 18 itself, but without success. Problem is that the latest version of polkadot/dev removes babel, and Im not sure how to replace that.

References https://github.com/webb-tools/relayer/issues/456


shekohex:
You will probably need to update the typegen too packages too similar to https://github.com/webb-tools/protocol-substrate/pull/319

Nutomic:
Okay based on that example I updated the package.json, removed everything related to babel, and ran typegen (which didnt generate any changes). Hope this works.

PR #593, opened 2023-05-02
[fix] Do not overwrite current signing round

1xstj:
**Summary of changes**
- When we have a large unsigned proposal queue, we would create a signing round every block, this could lead to a currently signing round being overwritten. This change adds a check to ensure that the current signing round will not be replaced if it is active and not stalled.

The flow would be 
1. Unsigned proposals found
2. Create signing sets
3. Creation fails since currently active signing round
4. https://github.com/webb-tools/dkg-substrate/blob/38f7b8387eff62b7587954e7c98466c31e89425d/dkg-gadget/src/worker.rs#L1612
5. Signing proposal queue is cleaned 
6. Repeat next block


dkg-gadget/src/worker.rs
```diff
@@ -416,10 +416,10 @@ where
 						lock[async_index as usize] = Some(status_handle)
 					} else if current_round.is_active() {
```
shekohex, line 417: As mentioned in our meeting, the `is_active()` will actually "mostly" return `true` since singing rounds stuck in `OfflineAndVoting` stage.

PR #577, opened 2023-04-18
[WIP] Signing Manager V2

tbraun96:
Resolves #580 
Resolves #539
Resolves https://github.com/webb-tools/dkg-substrate/issues/531
Resolves #602 
Part of #529

PR #606, opened 2023-05-15
Gossip Engine simplification

tbraun96:
The current implementation of the gossip engine has complexity that introduces more places for bugs to show up like messages not being delivered. For example, the use of sending notifications instead of the payloads through channels, the use of tokio::sync::broadcast (which can lag behind and lose messages entirely), and an acknowledgment system that does not take into account the state machine's processing result and is thus ineffective. We will need to build a better acknowledgment system later, and this PR helps clean up the rough edges that we currently have.

### DIFF
--- a/dkg-gadget/src/lib.rs
+++ b/dkg-gadget/src/lib.rs
@@ -174,8 +174,8 @@ where
 	keygen_gossip_engine.set_gossip_enabled(true);
 	signing_gossip_engine.set_gossip_enabled(true);
 
-	keygen_gossip_engine.set_processing_already_seen_messages_enabled(false);
-	signing_gossip_engine.set_processing_already_seen_messages_enabled(false);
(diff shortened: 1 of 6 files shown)


shekohex (commented):
LG!

dkg-gadget/src/lib.rs
```diff
@@ -174,8 +174,8 @@ where
 	keygen_gossip_engine.set_gossip_enabled(true);
 	signing_gossip_engine.set_gossip_enabled(true);
 
-	keygen_gossip_engine.set_processing_already_seen_messages_enabled(false);
-	signing_gossip_engine.set_processing_already_seen_messages_enabled(false);
+	// keygen_gossip_engine.set_processing_already_seen_messages_enabled(false);
```
drewstone, line 177: Do we need this? Why is it commented out?

1xstj:
Merged into #577

PR #607, opened 2023-05-17
Add shared key for CI Rust cache to speedup builds

Nutomic:
I noticed in the Slack chat that this repo also has problems with slow CI time. While solving the same problem in relayer I noticed that adding this shared-key for rust-cache helps a lot to speed up builds. So lets see if it helps here too.

**Summary of changes**
- Set shared-key for rust-cache CI action

Closes

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -45,6 +45,8 @@ jobs:
 
       - name: Rust Cache
         uses: Swatinem/rust-cache@v1.3.0
+        with:
+          shared-key: "rust"
 
       - name: Run Linters
         run: |
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -12,7 +12,7 @@ env:
   CARGO_REGISTRIES_CRATES_IO_PROTOCOL: git
 
 jobs:
-  # code coverage job; moved to own workflow file due to running out of disk space. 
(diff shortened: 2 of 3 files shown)


1xstj (approved):

Nutomic:
Unfortunately it didnt help much, Rust build time for e2e/integration tests is still 33 minutes just like before. This is because we are using release builds, and those have incremental compilation disabled by default. So there are two possibilities, either use debug builds for tests, or enable incremental compilation for release builds. The former is preferable because it should also give better debug info and other build speedups. Incremental release builds could create problems for production builds so I wouldnt do that.

tbraun96:
We need to take the latter approach. Building in release for testing is required due to performance reasons. Let's enable incremental compilation for release builds.

Nutomic:
In that case I would rather set the `opt-level` for debug builds.

https://doc.rust-lang.org/cargo/reference/profiles.html

1xstj:
The `e2e` tests fail with debug build

tbraun96:
You can append this to the file `node/standalone/Cargo.toml`

```toml
[profile.release]
incremental = true
```

tbraun96:
We need to use release builds for testing, not debug builds. That renders most of this PR, including its documentation, out of alignment with our methodologies.

PR #608, opened 2023-05-18
Enable incremental builds for release in workflows

tbraun96:
Incremental compilation is disabled by default for release builds. So, instead of using debug builds as was tried in #607 , we will instead keep release builds and set an env var to override incremental compilation.

To test this PR, we will first need the pipeline to finish once. Then, we will need to re-trigger (e.g., with an empty commit) the pipeline, and expect to see the rebuild time significantly shorter.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -31,7 +31,7 @@ jobs:
           echo "SCCACHE_GHA_ENABLED=true" >> $GITHUB_ENV
 
       - name: Run sccache-cache
-        uses: mozilla-actions/sccache-action@v0.0.2
+        uses: mozilla-actions/sccache-action@master
 
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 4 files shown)


tbraun96:
Looks like the `rust-cache` says:

```
... Cleaning /home/runner/work/dkg-substrate/dkg-substrate/target ...
... Cleaning cargo registry ...
... Cleaning cargo/bin ...
... Cleaning cargo git cache ...
... Saving cache ...
/usr/bin/tar --posix -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/dkg-substrate/dkg-substrate --files-from manifest.txt --use-compress-program zstdmt
Cache Size: ~1417 MB (1[48](https://github.com/webb-tools/dkg-substrate/actions/runs/5015346889/jobs/8990786612#step:27:49)5425[50](https://github.com/webb-tools/dkg-substrate/actions/runs/5015346889/jobs/8990786612#step:27:51)5 B)
Cache saved successfully
```

Will proceed to test to see if the cache is indeed used now.

1xstj (approved):

PR #610, opened 2023-05-19
ProposalSetUpdate and RefreshVote proposals should never expire

### LINKED ISSUE #609: [TASK] ProposalSetUpdate and RefreshVote proposals should never expire
**Task summary**
Currently all proposals have an expiry limit, but it would be better if we can assign that depending on the type of proposal, certain types namely `ProposalSetUpdate` and `RefreshVote` is necessary for keygen process to complete and session to rotate, these should never expire and should be signed.

1xstj:
**Summary of changes**
- keygen related proposals are never considered for expiry removal

Closes #609


1xstj:
replaced by #611

PR #611, opened 2023-05-19
[fix] skip expiration for keygen related proposals

### LINKED ISSUE #609: [TASK] ProposalSetUpdate and RefreshVote proposals should never expire

1xstj:
**Summary of changes**
- skip expiration for keygen related proposals

Closes #609

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -4,9 +4,6 @@ on:
   push:
     branches: [master]
   pull_request:
-    branches:
-      - "**"
-
   workflow_dispatch:
 
 env:
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -4,15 +4,13 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master ]
-
   workflow_dispatch:
 
 env:
   CARGO_REGISTRIES_CRATES_IO_PROTOCOL: git
 
 jobs:
-  # code coverage job; moved to own workflow file due to running out of disk space. 
+  # code coverage job; moved to own workflow file due to running out of disk space.
   # The runner will stop working when the machine runs out of disk space. Free space left: 72 MB
   coverage:
     name: coverage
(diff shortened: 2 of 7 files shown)


shekohex (commented):
Makes sense. But we also should not add more as long as we already have one queued, right?

tbraun96:
> Makes sense. But we also should not add more as long as we already have one queued, right?

Yes, this would make sense.

1xstj:
> Makes sense. But we also should not add more as long as we already have one queued, right?

Yes this will never happen though, so we run keygen the NextDkgKey is added to chain then a RefreshProposal is created. So lets say this is never signed, a new keygen will not run since the NextDKGKey is already onchain. This is what leads to us stalling since the RefreshProposal expired without signature and a new keygen wont execute.

shekohex:
> > Makes sense. But we also should not add more as long as we already have one queued, right?
> 
> Yes this will never happen though, so we run keygen the NextDkgKey is added to chain then a RefreshProposal is created. So lets say this is never signed, a new keygen will not run since the NextDKGKey is already onchain. This is what leads to us stalling since the RefreshProposal expired without signature and a new keygen wont execute.

Did not have a Auto Trigger or something like that which runs every `N = 10` blocks and see if we have a Next Public Key and that we do not have a Refresh Proposal and in that case it inserts a new one? If we still have that logic still, I would say we should remove it, as it is not useful now (after this PR)

1xstj:
> > > Makes sense. But we also should not add more as long as we already have one queued, right?
> > 
> > 
> > Yes this will never happen though, so we run keygen the NextDkgKey is added to chain then a RefreshProposal is created. So lets say this is never signed, a new keygen will not run since the NextDKGKey is already onchain. This is what leads to us stalling since the RefreshProposal expired without signature and a new keygen wont execute.
> 
> Did not have a Auto Trigger or something like that which runs every `N = 10` blocks and see if we have a Next Public Key and that we do not have a Refresh Proposal and in that case it inserts a new one? If we still have that logic still, I would say we should remove it, as it is not useful now (after this PR)

These are two different cases
1. Keygen fails : In that case, the auto trigger will tell the dkg-gadget to run the keygen process again after 10 blocks
2. Keygen suceeds and signing fails : In this case also the auto trigger will tell the dkg-gadget to run the keygen process again but the gadget sees that the NextDKGKey is already onchain so skips it

This is addressing point2

shekohex (approved):
LG!

PR #613, opened 2023-05-19
Experiment with mac runners for e2e tests

tbraun96:
Mac runners are free (for open-source projects), and, offer higher capabilities compared to the Linux runners we currently use (which means better performance). This may help make e2e tests more reliable in the pipeline, since flakiness is an issue for e2e tests.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -15,7 +15,7 @@ jobs:
   # dkg-substrate integration tests
   e2e:
     name: e2e-tests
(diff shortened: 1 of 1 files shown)


tbraun96:
Unexpectedly, the build took 2-3x as long. There does appear to be complaints about mac runners unexpectedly taking longer elsewhere, so I will close this.

PR #619, opened 2023-05-25
fix: pick oldest proposal to sign

### LINKED ISSUE #618: [TASK] Pick the oldest proposal to sign
**Task summary**
Currently the dkg-gadget looks at the unsigned queue and picks a proposal to sign. When we have a large number of unsigned proposals, this can lead to new proposals being signed while the old ones (nearing expiry) are not picked for signing.

Screenshot 2023-05-25 at 10 43 19 PM

We need to pick the oldest proposal for signing, this can be done by looking at the expiry. Every unsigned proposal has an expiry attached to it, so `sort(asc(expiry))` should return the oldest proposal to sign.

The benefits of this approach:
- Proposals are always signed in-order (wrt nonce and expiry)
- Less chance of older proposals expiring when new proposals are constantly added
- This will also help us in `batching` since this type of ordering will ensure all nodes are looking at the same set of proposals while picking batches.

1xstj:
**Summary of changes**
- Resolves #618 

Closes

### DIFF
--- a/dkg-gadget/src/signing_manager/mod.rs
+++ b/dkg-gadget/src/signing_manager/mod.rs
@@ -48,8 +48,6 @@ impl<B: Block, BE, C, GE> Clone for SigningManager<B, BE, C, GE> {
 	}
 }
 
-// 1 unsigned proposal per signing set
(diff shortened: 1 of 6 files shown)


shekohex (commented):
LGTM! 👍

PR #620, opened 2023-05-26
[WIP] Signing Manager V2: Fix message loss bug

tbraun96:


tbraun96:
The harness tests pass because it uses the in-memory gossip engine, which has been debugged for this branch. The real gossip engine has not yet been debugged, thus, this will need to be debugged before getting the pipeline to pass.

dkg-gadget/src/async_protocols/state_machine_wrapper.rs
```diff
@@ -117,23 +125,35 @@ where
 		if let Some(err) = result.as_ref().err() {
 			self.logger.error(format!("StateMachine error: {err:?}"));
 		} else {
+			let debug_after = format!("{:?}", self.sm);
+			// a message that is properly received should always increment the number of round
+			// messages recieved. If it doesn't, we log that here
+			if debug_before == debug_after {
```
shekohex, line 131: there is no `==` for the `sm`? do we have to print debug them to actually compare them? this sounds like a bad idea.

tbraun96: Yeah, this is a bad idea. It was just a quick way to see if a specific type of behavior was happening. It should be removed, since this behavior never happened anyways.

PR #634, opened 2023-06-12
Allow acceptable block ID range

tbraun96:


dkg-gadget/src/async_protocols/mod.rs
```diff
@@ -290,31 +290,31 @@ pub enum ProtocolType<MaxProposalLength: Get<u32> + Clone + Send + Sync + std::f
 		i: KeygenPartyId,
 		t: u16,
 		n: u16,
-		associated_block_id: Vec<u8>,
+		associated_block_id: u32,
```
drewstone, line 293: Let's use `u64` here instead of `u32`s for these values. `BlockNumber` in prod is a `u64` at least afaik.

drewstone (changes requested):

PR #654, opened 2023-06-23
[fix] Handle edge-cases in misbehaviour reports

1xstj:
**Summary of changes**
- Ensure the misbehaviour reports are against an existing authority
- Ensure authority is not already jailed
- Ensure we never reach t==n
- Ensure next_bext_authorities never drop below 2


drewstone (approved):
Awesome work @1xstj, LG

PR #652, opened 2023-06-20
Keygen Manager

tbraun96:

PR #659, opened 2023-07-03
Add retry_id to DKG messages, add keygen message enqueueing

tbraun96:
In the spirit of expeditiousness, while #652 is being debugged, we will add message enqueueing and additional checks to keygen messages in order to reduce the number of avenues from which stalling may occur.

### DIFF
--- a/dkg-gadget/src/gossip_messages/public_key_gossip.rs
+++ b/dkg-gadget/src/gossip_messages/public_key_gossip.rs
@@ -59,11 +59,11 @@ where
 			.logger
 			.debug(format!("SESSION {} | Received public key broadcast", msg.session_id));
 
-		let (is_main_round, is_genesis) = {
+		let is_main_round = {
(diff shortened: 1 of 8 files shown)


dkg-gadget/src/gossip_messages/public_key_gossip.rs
```diff
@@ -103,7 +104,7 @@ where
 			store_aggregated_public_keys::<B, C, BE, MaxProposalLength>(
 				&dkg_worker.backend,
 				&mut lock,
-				is_main_round,
+				is_genesis,
```
tbraun96, line 107: Bug fix. The function signature requires `is_genesis`, not `is_main_round`

shekohex (commented):
Should it be a `retry_count` instead of `retry_id`? an `id` stands for identifier, but I do not think we are using it for identifying anything? no?

PR #660, opened 2023-07-03
Merge refresh and proposer set proposals and unify more flows.

### LINKED ISSUE #635: [TASK] Use regular proposal submission flow for RefreshVote proposal
**Task summary**
Currently after being signed by the dkg-gadget the RefreshVotes are dealt seperately like below, https://github.com/webb-tools/dkg-substrate/blob/53622c613a951afd90e278ab429e75ba3e5fe76b/dkg-gadget/src/proposal.rs#L51 and picked up by the proposal handler like this https://github.com/webb-tools/dkg-substrate/blob/53622c613a951afd90e278ab429e75ba3e5fe76b/pallets/dkg-proposal-handler/src/lib.rs#L615

The task is to integrate this into the common proposal submission flow so that we can reduce a lot of redundant code

### LINKED ISSUE #651: [TASK] In the worst-case scenario, ensure we pick a single validator as the governor (use their DKGId)
**Issue summary**

We currently have an edge case where we don't reduce the `n` below 2. In reality, we should consider the case where everything fails and there is only 1 EOA signer remaining who becomes the next governor. This is good for liveness and then we can work to restart the keygen in the next round or sooner when this edge case happens.

Things that need to happen:
- Handling the case when we drop to `t=0` `n=1`
- DKGGadget would cease to execute normally and instead we would fallback to the single signer `keystore.sign(public, ...)`

drewstone:
**Summary of changes**
- Merges the `RefreshProposal` (prev `RefreshVote`) with the `ProposerSetUpdateProposal`.
- Removes separate `RefreshProposal` flow and unifies into signed proposal flow.
- Uses `OnSignedProposal` to handle any cleanup/update from signed `RefreshProposal` handling.
- Rename `GossipMsgPayload` to `NetworkMsgPayload` and remove `DKG` prefix from some messages.
- Move all network messages to `dkg-runtime-primitives`
- Adds `ProposerVote` to the repo which combined with `RefreshProposal` allows one to force change authorities/governor on Tangle cross-chain applications.
- Submits a vote for the force changed governor.

- Closes #651 (second comment on Refresh topic)
- Closes #635


drewstone:
Need your thoughts @1xstj, how we want to submit the votes for the fallback mechanism.

parachain/runtime/rococo/src/lib.rs
```diff
@@ -0,0 +1,1140 @@
+// Copyright 2022 Webb Technologies Inc.
```
1xstj, line 1: This file was removed, I think it got mixed up in the merge

pallets/dkg-metadata/src/lib.rs
```diff
@@ -1574,6 +1492,126 @@ impl<T: Config> Pallet<T> {
 		}
 	}
 
+	pub fn create_refresh_proposal(pub_key: Vec<u8>, nonce: u32) -> Proposal<T::MaxProposalLength> {
+		let uncompressed_pub_key = Self::decompress_public_key(pub_key).unwrap_or_default();
```
1xstj, line 1496: is default() safe here?

pallets/dkg-metadata/src/lib.rs
```diff
@@ -1574,6 +1492,126 @@ impl<T: Config> Pallet<T> {
...
+	/// - The merkle root of the ordered voter set's Ethereum addresses.
+	/// - The session length in milliseconds
+	/// - The # of proposers accumulated in the merkle root
+	/// - The nonce of the refresh proposal
+	/// - The public key of the next DKG's shared public key.
+	///
+	/// This data is intended to be used to update the voter set on other blockchains
+	/// that need a fallback mechanism when the DKG is not available or needs to be fixed or
+	/// changed.
+	fn create_voter_set_data() -> ([u8; 32], u64, u32) {
```
1xstj, line 1533: Can this return type be more readable pls. Just type alias to u64/u32 will help a lot here.

1xstj (approved):
LGTM, just minor suggestions

pallets/dkg-metadata/src/lib.rs
```diff
@@ -1574,6 +1492,126 @@ impl<T: Config> Pallet<T> {
...
+		let mut height = Self::get_voter_set_tree_height(voters.len());
+		while height > 0 {
+			curr_layer = Self::next_layer(curr_layer);
+			height -= 1;
+		}
+
+		let mut root = [0u8; 32];
+		root.copy_from_slice(&curr_layer[0][..]);
+		root
+	}
```
shekohex, line 1613: Could we use a proper merkle tree implementation instead of rolling out our custom code here?

pallets/dkg-proposals/src/lib.rs
```diff
@@ -227,43 +226,20 @@ pub mod pallet {
...
-	pub type Proposers<T: Config> = StorageMap<_, Blake2_128Concat, T::AccountId, bool, ValueQuery>;
+	pub type Proposers<T: Config> =
+		StorageValue<_, BoundedVec<T::AccountId, T::MaxProposers>, ValueQuery>;
 
-	/// Tracks current proposer set external accounts
-	/// Currently meant to store Ethereum compatible 64-bytes ECDSA public keys
+	/// Tracks current ECDSA voting keys for each validator
+	pub type VotingKey<T> = BoundedVec<u8, <T as Config>::VotingKeySize>;
+	pub type VotingKeyTuple<T> = (<T as frame_system::Config>::AccountId, VotingKey<T>);
+	pub type VoterList<T> = BoundedVec<VotingKeyTuple<T>, <T as Config>::MaxProposers>;
```
shekohex, line 238: Are these supposed to be storage?

pallets/dkg-proposals/src/lib.rs
```diff
@@ -299,8 +275,12 @@ pub mod pallet {
 		ChainWhitelisted { chain_id: TypedChainId },
 		/// Proposer added to set
 		ProposerAdded { proposer_id: T::AccountId },
+		/// Queued proposer added to set
+		QueuedProposerAdded { proposer_id: T::AccountId },
 		/// Proposer removed from set
 		ProposerRemoved { proposer_id: T::AccountId },
+		/// Queued proposer removed from set
+		QueuedProposerRemoved { proposer_id: T::AccountId },
```
shekohex, line 283: Could you explain what `Queued` means here?

shekohex:
Also need to fix and lint.

dkg-runtime-primitives/src/proposal.rs
```diff
@@ -30,52 +27,98 @@ pub use webb_proposals::{
 
 #[derive(Clone, RuntimeDebug, scale_info::TypeInfo)]
 pub struct RefreshProposal {
+	/// The merkle root of the voters (validators)
+	pub voter_merkle_root: [u8; 32],
+	/// The session length in milliseconds
+	pub session_length: u64,
```
tbraun96, line 33: If the session length is in milliseconds, do we need 64-bits to store that data, or, can we opt for something lower?

```suggestion
	pub session_length: u32,
```

dkg-runtime-primitives/src/proposal.rs
```diff
@@ -30,52 +27,98 @@ pub use webb_proposals::{
...
+	fn max_encoded_len() -> usize {
+		Self::LENGTH
+	}
+}
+
 impl Decode for RefreshProposal {
 	fn decode<I: codec::Input>(input: &mut I) -> Result<Self, codec::Error> {
 		const NONCE_LEN: usize = ProposalNonce::LENGTH;
-		let mut data = [0u8; NONCE_LEN + 64];
+		let mut data = [0u8; 32 + 8 + 4 + NONCE_LEN + 64];
```
tbraun96, line 73:
```suggestion
		let mut data = [0u8; Self::LENGTH];
```

dkg-runtime-primitives/src/proposal.rs
```diff
@@ -30,52 +27,98 @@ pub use webb_proposals::{
...
-		let mut ret = [0u8; NONCE_LEN + 64];
+		let mut ret = [0u8; 32 + 8 + 4 + NONCE_LEN + 64];
+		let voter_merkle_root = self.voter_merkle_root;
+		let session_length = self.session_length.to_be_bytes();
+		let voter_count = self.voter_count.to_be_bytes();
 		let nonce = self.nonce.to_bytes();
-		ret[0..NONCE_LEN].copy_from_slice(&nonce);
-		ret[NONCE_LEN..(NONCE_LEN + 64)].copy_from_slice(&self.pub_key);
+		let pub_key = self.pub_key.as_slice();
+		ret[0..32].copy_from_slice(&voter_merkle_root);
```
tbraun96, line 107: For lines 107-111, we are copying bytes to an array. Finally, on line 112, we are heap allocating those bytes behind a `Vec`. This would perform a memcpy behind the scenes, which is not necessary in this case. Instead of declaring an array, you can declare a `vec![0u8; Self::LENGTH]` which already has a zeroed-out vec. Thus, we copy the bytes directly to where they need to in memory without a conversion. When finished, you just return that vec.

dkg-runtime-primitives/src/proposal.rs
```diff
@@ -30,52 +27,98 @@ pub use webb_proposals::{
...
+		ret[40..44].copy_from_slice(&voter_count);
+		ret[44..(44 + NONCE_LEN)].copy_from_slice(&nonce);
+		ret[(44 + NONCE_LEN)..].copy_from_slice(pub_key);
 		ret.into()
 	}
 
 	fn encoded_size(&self) -> usize {
 		const NONCE_LEN: usize = ProposalNonce::LENGTH;
-		NONCE_LEN + 64
+		32 + 8 + 4 + NONCE_LEN + 64
```
tbraun96, line 117:
```suggestion
		Self::LENGTH
```

pallets/dkg-metadata/src/mock.rs
```diff
@@ -126,14 +128,32 @@ where
 	}
 }
 
+pub const MILLISECS_PER_BLOCK: u64 = 10000;
```
tbraun96, line 131: Earlier, I left a comment about whether or not we need 64 bits to store the number of milliseconds. If we don't want to have to change the types here just to accommodate that, it's not a big deal.

tbraun96 (commented):
A few comments, and a few nits. Overall, this is a solid PR that will help improve the DKG.

dkg-runtime-primitives/src/proposal.rs
tbraun96: Related: https://doc.rust-lang.org/beta/src/alloc/vec/mod.rs.html#3119

pallets/dkg-metadata/src/lib.rs
drewstone: Yep will implement. Thanks for the feedback.

pallets/dkg-metadata/src/lib.rs
drewstone: I support this as a separate task.

pallets/dkg-proposals/src/lib.rs
drewstone: No they're for clippy to simplify the complex type.

pallets/dkg-proposals/src/lib.rs
drewstone: Sorry this can be removed, artifact from previous attempt. Thank you for spotting.

dkg-runtime-primitives/src/proposal.rs
drewstone: I think that `u64` is good. With a `u32` we are limited to a maximum session length of around 49-50 days I believe. Calculated as $50 * 3600 * 24 * 1000 = 4320000000 > 4294967296$

dkg-runtime-primitives/src/proposal.rs
drewstone: Thank you for the feedback I will update.

pallets/dkg-metadata/src/mock.rs
drewstone: I think let's leave this as is currently.

PR #663, opened 2023-07-05
Add type change checking

### LINKED ISSUE #662: [TASK] Add a github action to warn when types have changed
**Task summary**
Create a github action that runs 
```
yarn update:metadata
yarn build:interfaces
```

and generates the new types on every PR, if this creates a diff that is not part of the PR aka the types have changed but type definition has not been updated, the action should fail.

tbraun96:
Closes #662


1xstj:
Not needed after #669

PR #669, opened 2023-07-08
Update events for ease of relayer operations

drewstone:
**Summary of changes**
- Removes uncompressed pubkey from most events since it's unecessary and pushes more work into the runtime. Instead we should let front-ends/relayers decompress pubkeys as need be.
- Updates the events so that we emit the full refresh proposal that is signed, this way relayers can easily relay that instead of query and build any needed data.


drewstone:
Purpose of this is to update events so transfer process is more seamless for the relayer. We will need to update `webb-rs` this week @shekohex @1xstj including types. Let's plan to get the new types for these repos also published this week please.

drewstone:
This also deletes `dkg-test-suite`. If you object @1xstj @shekohex @tbraun96 please share thoughts. It seems that we spend more time updating and wrangling types to get those tests to work and I'm unconvinced that this is productive and that these tests are needed. They can just be pushed into Solidity unit tests. If we want proper E2E tests let's use Rust moving forward. The webb.js types system is not worth maintaining and we should move all useful libraries there to the `protocol-solidity` repo.

1xstj (approved):
I agree with these changes, will wait for @shekohex to review before merge

shekohex (approved):
LG

PR #676, opened 2023-07-13
Generalize the DKGs via modularization

### LINKED ISSUE #677: Generalize the DKGs
In the future, multiple DKGs will be used within the gadget. To prepare for this, we need to modularize our DKGs to make them selectable at runtime.

tbraun96:
Closes #677


drewstone (approved):
LGTM, this is entirely backwards compatible right?

tbraun96:
> LGTM, this is entirely backwards compatible right?

Yes, this will work with our current system. We never trigger the newer system since it's left unimplemented and never used.

PR #680, opened 2023-07-14
Generate multiple signing sets, create signing set IDs

tbraun96:
In an effort to increase the probability that signing a proposal succeeds in the event a signer drops off in a signing set, we now spawn 2 signing sets per unsigned proposal. Before SigningManagerV2, we did something similar, except, we did so in response to messages going missing. Now that messages are no longer mysteriously dropping, so long as nodes stay online and are non-malicious, we increase the probability that we converge toward signing completion.

### DIFF
--- a/dkg-gadget/src/signing_manager/mod.rs
+++ b/dkg-gadget/src/signing_manager/mod.rs
@@ -59,6 +59,7 @@ const MAX_RUNNING_TASKS: usize = 4;
 const MAX_ENQUEUED_TASKS: usize = 20;
 // How often to poll the jobs to check completion status
 const JOB_POLL_INTERVAL_IN_MILLISECONDS: u64 = 500;
+pub const MAX_SIGNING_SETS_PER_PROPOSAL: u8 = 2;
 
 impl<B, BE, C, GE> SigningManager<B, BE, C, GE>
 where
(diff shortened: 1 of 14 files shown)


dkg-gadget/src/signing_manager/mod.rs
```diff
@@ -73,8 +74,8 @@ where
 			work_manager: WorkManager::<B>::new(
 				logger,
 				clock,
-				MAX_RUNNING_TASKS,
-				MAX_ENQUEUED_TASKS,
+				MAX_RUNNING_TASKS * MAX_SIGNING_SETS_PER_PROPOSAL as usize,
+				MAX_ENQUEUED_TASKS * MAX_SIGNING_SETS_PER_PROPOSAL as usize,
```
drewstone, line 78: Seems dangerous given our previous issues with the size of the queue no?

drewstone:
Do we intend to merge this @1xstj @tbraun96 ? Let's get it in first thing if so.

dkg-gadget/src/signing_manager/mod.rs
tbraun96: The second implementation, which is after this commit, no longer includes this code.

drewstone:
Partially completed: https://github.com/webb-tools/tangle/issues/207

1xstj (approved):

PR #689, opened 2023-07-19
[WIP] improve signing sets

### LINKED ISSUE #688: [TASK] Blame Manager
When we obtain the `blame` for any given running protocol, we are given a `Vec<u16>`, which is a list of parties who have not been able to properly send a message to us. These indexes may not be the same between keygen and signing stages, since, for keygen, we generate the party indexes using either `dkg_worker.get_party_index()` or `dkg_worker.get_next_party_index()`, and for signing, we generate the party indexes using `dkg_worker.get_party_index()`.

Since we want to be able to generate signing sets for signing protocols based on the set of best authorities minus any blamed nodes (i.e., "wildcard signing sets"), we need an organized way of keeping track of blame during and between sessions while ensuring we don't mix up blame indexes. Additionally, we want to remove blamed nodes from this tracked list if they behave properly after not behaving correctly (e.g., due to a spurious network error).

The `BlameManager` will help us keep track of blamed party indexes for both signing and keygen. These indexes should be saved to a `HashMap<SessionId=u64, HashSet<PartyID=u16>>` that way we can track the blame for each with respect to a session ID.

Additionally, we want the `BlameManager` to be able to `filter` signing sets, replacing blamed nodes with properly-behaving nodes:

```rust
fn filter_signing_set(&self, input: &Vec<u16>, all_parties: &Vec<u16>) -> Vec<u16> {
    // TODO: replace any blamed nodes inside `input` with non-blamed nodes inside `all_parties`
}
```

We should keep track of blame for the current session, and, remove blame from past sessions each time we rotate sessions to keep the hashmaps cleared:

```rust
fn on_session_rotated(&self, new_session: u64) {
    // TODO: clear both hashmaps
}
```

By abstracting this away inside another manager, we further compartmentalize our code and reduce the chances we introduce bugs in the future when working between blames for keygen and signing party indexes. Additionally, this gives us a clean way to produce filtered signing sets that exclude blamed nodes.

tbraun96:


dkg-gadget/src/signing_manager/mod.rs
```diff
@@ -203,84 +206,182 @@ where
...
+
+				// If this is the last signing set, then, we should add any wildcard sets
+				// Definition of "wildcard set": a signing set that may or may not be symmetric
+				// to other generated signing sets since we exclude blamed part indexes from the
+				// generated signing set and replace with non-blamed parties. Since blames are not
+				// necessarily symmetric between nodes, it follows that these wildcard sets are
+				// not necessarily symmetric either. As such, the only way a wildcard set converges
+				// successfully to completion is if the same node(s) is blamed on all nodes
+				// participating in the signing protocol.
+				if ssid == MAX_POTENTIAL_SIGNING_SETS_PER_PROPOSAL - 1 {
```
1xstj, line 233: imo this blame logic will fire if any node is slow to respond too. Right now we have an elegant flow to enqueue those messages, but this blaming could make the situation worse, by starting t+1 asymmetric signing sets.

drewstone:
Is this still WIP @tbraun96 ?

drewstone:
Would be good to get this into the pre test net @1xstj

PR #690, opened 2023-07-20
[fix] improvements to proposal handling

1xstj:
**Summary of changes**
- Split batch creation and proposal expiry to seperate jobs, batch creation is first priority and expiration will run only with leftover weights
- Add ext to force_remove_proposal from queue
- Modify ProposalBatchRemoved and ProposalBatchExpired events
- Unify all dkg-gadget constants to constants.rs file
- Reduce signing_manager `MAX_RUNNING_TASKS` to `1`


shekohex (approved):
Looks good!

PR #694, opened 2023-07-25
Add slashing/offences on submission of signed non existent proposals/malformed proposals

### LINKED ISSUE #537: [TASK] Add slashing / offences to presenting signatures that never had proposals submitted
**Issue summary**
We have the property that we only sign proposals that land on chain. This doesn't mean that the nodes can't sign anything they want. In that case, we should offend and slash them. I think we can leverage the existing `offences` pallet to make this easy. Namely, when this happens we should have an extrinsic to submit this signature, prove it wasn't on-chain, and then add offenses to all the authorities.

1xstj:
**Summary of changes**
Resolves #537


shekohex (approved):
Looks good! but just a random question, who would report these to the chain? is it the other nodes? relayers? anyone else?

pallets/dkg-proposal-handler/src/mock.rs
```diff
@@ -169,6 +175,26 @@ where
 	}
 }
 
+type IdentificationTuple = (AccountId, AccountId);
+type Offence = crate::DKGMisbehaviourOffence<IdentificationTuple>;
+
+parameter_types! {
+	pub static Offences: Vec<(Vec<AccountId>, Offence)> = vec![];
```
tbraun96, line 182: Would it maybe make more sense to use:

```HashMap<Offence, Vec<AccountId>>```

File: pallets/dkg-proposals/src/mock.rs [hunk 2]
```diff
@@ -248,6 +259,26 @@ impl pallet_collator_selection::Config for Test {
 	type WeightInfo = ();
 }
 
+type IdentificationTuple = (AccountId, AccountId);
+type Offence = pallet_dkg_proposal_handler::DKGMisbehaviourOffence<IdentificationTuple>;
+
+parameter_types! {
+	pub static Offences: Vec<(Vec<AccountId>, Offence)> = vec![];
```
tbraun96, line 266: Same here

pallets/dkg-proposal-handler/src/mock.rs
1xstj: Hashmap is not recommended in substrate no_std env
https://substrate.stackexchange.com/questions/6929/can-i-store-a-hashmap-in-storagemap

PR #695, opened 2023-07-31
SBP Milestone review 3

hbulgarini:
The code reflects a team very experienced with something development and there are not so many action items to point out (at least from my point of view).

All the previous items pointed out in the SBP Milestone review 2 were addressed. 

I have only spotted one issue that should be addressed which is the extrinsic `force_reset_indices`. Please refer to the comments in the PR to more details.


1xstj:
Thankyou, opened issue https://github.com/webb-tools/dkg-substrate/issues/696 to track

PR #706, opened 2023-09-07
Merge Gossip Engine, Add Rust-based integration tests

### LINKED ISSUE #693: Merge the split keygen/signing gossip engines
The gossip engines were earlier split to help us debug issues in the gossiping engine while having message loss problems. Now that the message loss problems are gone, we should only have a single gossiping manager. This will also help us ensure consistency when eventually implementing the network topology aware signing sets.

tbraun96:
Run the integration tests with`cargo run --bin dkg-integration-tests`

Resolves #693

### DIFF
--- a/.github/workflows/harness_stress_tests.yml
+++ b/.github/workflows/harness_stress_tests.yml
@@ -1,4 +1,4 @@
-name: harness stress tests
+name: Stress Tests
 
 on:
   push:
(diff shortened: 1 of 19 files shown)


drewstone:
Is this still WIP?

drewstone (approved):
LG

PR #709, opened 2023-09-08
fix: update reputation logic

### LINKED ISSUE #707: [BUG] DKG reputations query does not return reputation of all validators
### Overview

`dkg.authorityReputations()` query only returns reputation of `4` validators even though there are `5` validators present. We need to return reputation values for all the available validators.

CleanShot 2023-09-08 at 04 32 33

1xstj:
**Summary of changes**
- Increase reputation on submission of next public key

Closes #707


salman01zp (approved):
LG..!!

drewstone (approved):

PR #711, opened 2023-09-15
Add WT-FROST to the worker gadget

### LINKED ISSUE #710: Implement FROST DKG into the DKG
[We currently have a PR up](https://github.com/Trust-Machines/stacks-sbtc/pull/697) on a FROST-related repo that will enable us to use that repo properly with the DKG gadget. We can use the fork, then, we can use the official repository once the PR is merged to add a `DKGModule` for FROST.

**Steps**

* Prototype the protocol using a basic network test engine, sort out any nuances (https://github.com/webb-tools/dkg-substrate/issues/712)
* Add a `DKGModule` trait implementation for FROST, e.g., `FrostModule` (https://github.com/webb-tools/dkg-substrate/issues/713)
* Add on-pallet logic for unsigned proposals and keygens (i.e., should_execute_keygen related information) that uses a single byte-sized field for denoting which `DKGModule` to use for the given unsigned proposal or keygen (https://github.com/webb-tools/dkg-substrate/issues/714)
* Determine a mechanism for selecting a `DKGModule` during runtime (https://github.com/webb-tools/dkg-substrate/issues/714)
* Add integration test that selects the `FrostModule` and runs it in the local chain environment https://github.com/webb-tools/dkg-substrate/issues/715

tbraun96:
Closes #710


drewstone (changes requested):
CI needs to be fixed.

drewstone:
Maybe you're missing some feature flags @tbraun96 ?  What ultimately changed with how our integration tests run?

tbraun96:
> What ultimately changed with how our integration tests run?

So far, nothing changed in how our integration tests are run. Relevantly, what has changed is that our FROST dependency needs to link to a C library, and it appears it lacks testing for pipeline builds (`cargo test` does not count, as the linking process is different. What is failing for us here is when we build binaries via `cargo run` or `cargo build`). I am in contact with @xoloki about this situation and am exploring ways to fix this issue. We may need a custom build.rs script for our frost dependency.

xoloki:
I'm really not sure what's going on here.   In [stacks-sbtc](github.com/Trust-Machines/stacks-sbtc) we link `wsts` with the `p256k1` dependency, and CI works fine on `x86_64` and `arm64` for both `linux` and `macos`.  All we needed to do was add `libclang-dev` to our build pipeline.

I'll pull this repo and try to see what's going on.  `p256k1` has its own `build.rs`, which should be adequate on all platforms.

xoloki:
> I'll pull this repo and try to see what's going on. `p256k1` has its own `build.rs`, which should be adequate on all platforms.

Just pulled and built `thomas/frost-dkg`, no problems.  I just had to install `protobuf-compiler` and `libgmp-dev`.

drewstone:
Just in CI @xoloki

xoloki:
@drewstone @tbraun96 there seems to be a problem in how CI is setup and run.  

I would expect that the `Harness` CI target comes from `.github/workflows/harness_stress_tests.yml`.  But in that file, the `Install Protobuf and LLVM` step should run `bitcoin.sh` then `sudo apt-get install protobuf-compiler llvm libsecp256k1-dev libsecp256k1-0 libclang-dev`.  

However, that's not what happens.  The `bitcoin.sh` script does not run, and `apt-get` is run twice, once for `protobuf-compiler` and once for `llvm`.

xoloki:
However, for the `Integration tests / Local Chain Tests` target, we are actually installing `libclang-dev`, but the linker problem persists.  

There are a couple of things about your CI setup which I note.  First, you are using `sccache` for both failing tests.  Second, you are using `cargo-make` for the integration test.  Maybe remove these and try a basic `cargo build` step in each?

xoloki:
I'd also remove the `libsecp256k1` install calls, unless you're also using those for something else.  

Basically, I'd just remove everything in the failing CI `.yml` files and build them from scratch, installing only what you need and seeing if `cargo build` works.  Once that works then you can add stuff back in one piece at a time.

PR #725, opened 2023-10-24
fix : Accepting a RefreshProposal should clear all pending unsigned RefreshProposals

### LINKED ISSUE #724: [BUG] Accepting a RefreshProposal should clear all pending unsigned RefreshProposals
**Describe the bug**
Consider the case where we have a DKG stall on session 99, we execute force_rotate, this results in two RefreshProposals in UnsignedQueue 

```
Unsigned Queue 
 - RefreshProposal : Nonce 100
 - RefreshProposal : Nonce 101
```

Once the DKG starts signing again, it will sign both RefreshProposals and end up with two session rotations.

**Expected Behaviour**
With multiple refresh proposals in queue, only one should be signed/accepted

1xstj:
**Summary of changes**
- Resolves #724


drewstone (commented):
LG just need to fix CI

drewstone (approved):

